PR #600, opened 2025-01-24
feat: optimise normalisation over binary values

DavePearce:
For a normalisation `(~ X)` arises for a binary column `X` (or expression resulting in a binary value) then, in fact, no normalisation is actually required.  Instead, we can just return `X` since it is already normalised.  This is a useful optimisation since it will save an inverse column, along with two constraints.


OlivierBBB:
Was this feature not already in `corset` ? Just for completeness there is an associated optimization for `if ... then ... else ...` constraints, when the condition is binary. I did notice a while back that we were producing a lot of inverses that didn't need to exist.

PR #695, opened 2025-02-14
fix: incorrect branches calculation

Fallengirl:
https://github.com/Consensys/go-corset/issues/686


DavePearce:
Not actually a complete solution, but headed in the right direction.

PR #715, opened 2025-02-21
feat: lexicograpgic ordering constraint

DavePearce:
adds support for the `defsorted` construct.


DavePearce:
@letypequividelespoubelles FYI

PR #723, opened 2025-02-24
feat: strict lexicographic ordering constraint

DavePearce:


DavePearce:
@letypequividelespoubelles FYI

Issue #601, opened 2025-01-24
feat: reject input columns without types


DavePearce:
The goal here is to require that all input columns are given explicit columns.  Thus, a constraint file containing something like this would be rejected:

```
(defcolumn X Y)
```

This is because neither `X` nor `Y` have types.  Instead, we could write something like this:

```
(defcolumn (X :i16) (Y :i16))
```

Before this can be implemented, however, there are some checks and actions required:

1. (**Query**) Firstly, we need to consider what columns existing within the constraints that do not have explicit columns, and why not.
2. (**Action**) Secondly, we need to update the existing constraints to ensure that all input columns have explicit types.

Once these two tasks are done, this restriction can be implemented whilst ensuring the existing constraints continue to compile.

amkCha:
**1. (Query)**

As of the latest constraints with commit `81d5c3a`, the following 3 columns do not have explicit types : 
- `inst_decoder.OPCODE` (module `reftables`) 
```
 (OPCODE :display :opcode)
```
- `loginfo.TOPIC_HI`
- `loginfo.TOPIC_LO`
```
  (TOPIC_HI :array [4])
  (TOPIC_LO :array [4])
```
Why ? for the first one, it looks like `:byte` was forgotten. For the last two, it gives the impression that it's typed because it's defined as an array. To avoid omissions of the type, we could change it to typed arrays in `go-corset` ? 
Found by running on this branch https://github.com/Consensys/go-corset/tree/feat/field-type-column-from-bin
```
go run main.go query --field-columns zkevm.bin
```
Test
- Added a dummy module with untyped columns and columns were printed in the output

DavePearce:
Great, those are the three columns I found also.  So, that's good news --- not too many!

> for the first one, it looks like :byte was forgotten

Right, so we can check with the team and update it in `linea-constraints`.

> To avoid omissions of the type, we could change it to typed arrays in 

Well, we have to be careful here.  We must change `linea-constraints` before we change `go-corset`, otherwise we end up in a position where the current constraints cannot be compiled.  Therefore, the first step is to add the right type to `linea-constraints` for these two columns `TOPIC_HI` and `TOPIC_LO`.  I actually don't know what type they wanted, so we need to figure that out.

Once those updates to `linea-constraints` are merged in, then we can think to update `go-corset` so that it rejects all column declarations without explicit types.

DavePearce:
Also, looked quickly at your changes.  I didn't understand what you're trying to do here:

```
[...]

PR #758, opened 2025-03-10
feat: reject columns no type (#601)

DavePearce:


pkg/corset/compiler/parser.go
```diff
@@ -389,11 +390,25 @@ func (p *Parser) parseColumnDeclaration(context util.Path, path util.Path, compu
...
-	} else {
+	} else if computed {
+		// Only computed columns can be given without attributes.
 		name = *path.Extend(e.String(false))
+	} else {
+		return nil, p.translator.SyntaxError(e, "column is untyped")
+	}
+	// Final sanity checks
+	if computed && datatype == nil {
+		// computed columns initially have multiplier 0 in order to signal that
```
amkCha, line 404: From the if/else if above, I understand that when `computed`, we set the name and not the datatype, so `datatype` is always `nil` no ? So I'm sure I get why we have a case where we set it to field type and another when we error that "computed columns cannot be typed" :thinking:

pkg/test/valid_corset_test.go
```diff
@@ -1178,10 +1177,6 @@ func Test_Native_11(t *testing.T) {
 // Equivalence Tests
 // ===================================================================
 
-func Test_Equiv_01(t *testing.T) {
```
amkCha, line 1181: We remove that test on purpose ?

pkg/corset/compiler/parser.go
DavePearce: Hmmm, so here's an example where the error `computed columns cannot be typed` is reported:

```
(defcolumns (X :i16) (Y :i16))
(defpermutation ((Z :i16)) ((+ X)))
```

In principle, this is actually OK.  However, I just haven't implement logic to handle computed columns with explicit types.  At the moment, the type of a computed column is always inferred and this overwrites any initial type given.  So, in the above, the type given to `Z` makes no difference as it is overwritten by the type of `X` (which is `i16`).  

So, I think the answer is: for now, we handle the case for a computed column where no type is given; but, for the case where a type is given, we need to implement some additional logic (at some point).

pkg/test/valid_corset_test.go
DavePearce: Good spotting!!! I just renamed it from `equiv_01` to `stdlib_03`.  There's no really good reason for me doing this, I just felt it was better to consolidate these two test groups together.  This is actually not relevant to this PR, but was something I thought about doing before and just did it here.

PR #984, opened 2025-07-08
Fix: small parsing bug in https://github.com/Consensys/go-corset/pull/980

Tabaie:


pkg/util/assert/assert.go
```diff
@@ -77,3 +82,108 @@ func asInt64(x any) (int64, bool) {
 
 	return 0, false
 }
+
```
DavePearce, line 85: Using `callerInfo` seems like overkill to me.  And, I don't think we should take stuff some `testify`.  Useful for debugging, but now tests are passing ...

DavePearce (commented):
Yeah, not sure about using `callerInfo` here taken from `testify`.  Seems like debugging code to me.

pkg/util/assert/assert.go
DavePearce: I hear you.  But actually, this stuff is not going to fail once you've got it working.  You also won't necessarily need all of this to debug the problem.  In most cases it will be unnecessary.

DavePearce (approved):
LGTM

PR #994, opened 2025-07-10
feat: some small field operations

Tabaie:
A basic implementation of simple (small) field operations.

* It uses a unified element type and a common field implementation. The elements are passive object, with operations being methods of the field.
  - Advantage: Flexible, conducive to easy and quick changing of the field.
  - Advantage: Readable, not requiring code-generation.
  - Disadvantage: More similar to the `api` objects in gnark than in the field elements in gnark-crypto, making a unified small/large field interface a bit difficult. (My instinct is to optimize the interface for this case and make the large field implementation conform to it. Since a small-field migration is inevitable, a moderate performance penalty should be acceptable.)

* Different implementation from gnark
  - Advantage: Enables small-field-native common sense interface features such as passing elements by value rather than reference.
  - Disadvantage: gnark is widely tested and optimized, including assembly implementations for both x64 and ARM.


Tabaie:
Closing in favor of https://github.com/Consensys/go-corset/pull/999

PR #999, opened 2025-07-12
feat: Small Fields, Interface, and Wrapper for the bls12-377 Scalar Field

Tabaie:
* Purego implementations for small fields of interest (Mersenne31, Koalabear, GF251 and GF8209)
* An interface that the small field impls natively support
* A wrapper for the BLS12-377 scalar field, also implementing the interface.


field/bls12-377/element.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// AddUint32 x + y. It's the canonical way to create new elements.
+func (x Element) AddUint32(y uint32) Element {
+	res := fr.NewElement(uint64(y))
+	return Element{res.Add(x.Element, &res)}
+}
+
+// ToUint32 returns the numerical value of x.
+func (x Element) ToUint32() uint32 {
```
DavePearce, line 32: This is fine, but we will need methods to allow creating a field element from `[]byte` and writing out a field element to `[]byte` (in big endian I suppose to match gnark for now).

field/interface.go
```diff
@@ -0,0 +1,18 @@
+package field
+
+import "fmt"
+
+// An Element of a prime-order field.
+type Element[Operand any] interface {
```
DavePearce, line 6: I think we need a few more methods to allow reading from `[]byte` and possibly also writing back to `[]byte`.

DavePearce: Also, I think `CmpUint32` or similar would be good since I do a lot of comparisons against `0` and `1`, and possibly other values.

field/gf251/element_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/consensys/go-corset/pkg/util/assert"
+)
+
+func TestMul(t *testing.T) {
+	var i, j, m big.Int
+
+	m.SetUint64(modulus)
+
+	for range 10000 {
```
DavePearce, line 32: I would suggest a different testing strategy in this case.  It should be possible to enumerate **all combinations**, and I would do this.

field/gf251/element.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return Element{res}
+}
+
+// montgomeryReduce x -> x.R⁻¹ (mod m)
+func montgomeryReduce(x uint64) Element {
+	// textbook Montgomery reduction
+	m := (x * uint64(negModulusInvModR)) % r // m = x * (-modulus⁻¹) (mod r)
+
+	res := uint32((x + m*modulus) / r)
```
DavePearce, line 61: I seem confused about this.  Are you just working on the assumption that the compiler will turn `/ r` into `>> 5` ?

field/gf251/element.go
```diff
@@ -0,0 +1,160 @@
...
+	"cmp"
+	"math/bits"
+	"strconv"
+)
+
+// Element of a prime order field, represented in Montgomery form to speed up multiplications.
+type Element [1]uint32 // defined as an array to prevent mistaken use of arithmetic operators, or naive assignments.
+
+const (
+	r                 = 1 << 32 // register size
```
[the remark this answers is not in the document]
Tabaie: Yes we can choose to define element as `[1]uint16` instead, in which case `r` will be `65536`.

field/interface.go
Tabaie: If comparing with 0 and 1 are done frequently how's dedicated `IsZero()` and `IsOne()` methods?

field/gf251/element.go
Tabaie: Yes I'm assuming division and modulus operations against a constant power of two is optimized into shifts and ands.

field/bls12-377/element.go
Tabaie: Made methods `Bytes` and `AddBytes`.

DavePearce (approved):
LGTM

PR #1057, opened 2025-07-29
refactor: `Trace` -> `Trace[field]`

Tabaie:
This PR generifies the `Trace` interface, but keeps all its instantiations to `bls12_377` as before, to facilitate further refactoring and an eventual migration to other fields.


DavePearce (approved):
LGTM --- though I thought you were going to make `trace.Column` generic as well?  Anyway, can do in a subsequent PR no problem.

PR #1082, opened 2025-08-05
feat: Batch Inversion

Tabaie:
This PR adds

* a generic BatchInvert function
* some comments documenting bit length invariants in each small field subpackage

NOTE: Each invocation of `BatchInvert` creates a scratch slice of the same size as its input slice. It may be desirable to use a pool for this (or even have it as a method of the pool).


pkg/util/field/field_array.go
```diff
@@ -165,3 +165,76 @@ func (p *FrElementArray) String() string {
...
+
+	s[len(s)-1] = inv
+	if isZero.get(len(s) - 1) {
+		s[len(s)-1] = zero
+	}
+}
+
+// boolSlice compactly represents a boolean slice, low bytes and bits first.
+// It does not strictly enforce its size.
+type boolSlice struct {
```
DavePearce, line 219: I think you can use an existing type `util/collection/bit/bit_set.go` for this.  You might like to add a `Set(bool)` method to it, that's fine.

Tabaie: Great! I'll use that instead.

PR #1119, opened 2025-08-19
Feat/basic field arith with asm

amkCha:


testdata/asm/gf251_mul.accepts
DavePearce, line 1: I think we need to sample this file down to say `5000` rows or something.  Example command I would use for this:

```
> shuf gf251_mul.accepts | tail -n 5000 > gf251_mul.sampled.accepts
```

testdata/asm/gf251_mul.zkasm
```diff
@@ -0,0 +1,57 @@
...
+
+   var quot u8
+   var tmp_quot u16
+   var quot_times_n u16
+   var discard_bit_shift u8
+   var res u8
+   var red u8
+   var c0, c1 u1
+
+   ninv_mod_R = 205
```
DavePearce, line 23: Its not good to have registers used to hold constant values (as these create additional columns in the trace).  Better to just use the values directly for now.  At some point, we'll have the ability to create named constants and we can apply them here.  Leave a comment perhaps to clarify though what each constant is.

testdata/asm/gf251_add.zkasm
```diff
@@ -0,0 +1,22 @@
+;; GF251 curve
+;; Add on u256
+;; Mul on u8, to be able to test all cases
+
+fn add(x u256, y u256) -> (RESULT u256) {
```
DavePearce, line 5: I don't understand why you have `add(u256,u256) -> u256`.  Should be `add(u8,u8)->u8`.

testdata/asm/gf251_mul.zkasm
```diff
@@ -0,0 +1,57 @@
...
+   if c0==0 goto exit_1
+   RESULT = res
+   return
+exit_1:
+   RESULT = red
+   return
+}
+
+;; TODO: uncomment once fn call in fn is fixed
+;;fn mul_u8(x u8, y u8) -> (RESULT u8) {
```
DavePearce, line 52: Don't think you need to call it `mul_u8` and `reduction_u8`.  Seems like `mul` and `reduction` is fine?  Anyway, not exactly a big deal.

testdata/asm/gf251_add.accepts
DavePearce, line 1: Hmmmm, so we not want more tests here?

PR #1149, opened 2025-09-03
perf: Integrate Batch Inversion in the PseudoInverse gadget

Tabaie:
This PR adds a `PseudoInverse` assignment to compute for vector $[v_i]$, the vector $[w_i]$ where $w_i = v_i^{-1}$ if $v_i = 0$ and $0$ otherwise. The computation has been accelerated by using batch inversion.


DavePearce (approved):
LGTM

PR #1552, opened 2026-03-05
1537 add support type for aliases

### LINKED ISSUE #1537: feat: add support for type aliases to ZkC
## Summary

The goal of this issue is to add support for "type aliases" to the ZkC language which, in particular, will be useful when `struct` types are added later (see #1543).  For example:

```
type address = u160

input storage(account address, location u256) -> (value u256)

function read(account address, location u256) -> (value u256) {
   return storage[account,location]
}
```

## Approach

This is a moderately challenging issue to implement, and has many similarities with the existing mechanism for constants (see `zkc/compiler/ast/decl/constant.go`).  Some specific challenges:

1. **DataType**. A new type must be added to `zkc/compiler/ast/data/` for "type aliases".  This will force a non-trivial change of `data.Type` into `data.Type[S symbol.Symbol]` which needs to be propagated throughout the compiler.
2. **Declaration**.  A new declaration form (`TypeAlias`?) must be added to `zkc/compiler/decl`, and the parser should be updated to parse the syntax and constructs an instance of a `TypeAlias` declaration.
3. **Resolution**.  The name resolution procedure (currently defined as the poorly named `linker.Link()` function) needs to be updated to resolve types from `Type[symbol.Unresolved]` to `Type[symbol.Resolved]` (very roughly following how thing are done for e.g. constants)
4. **Typng**.  The `validate.Typing()` function needs to be updated to handle type aliases when it encounters them.  Furthermore, it also needs a check to prevent cyclic type aliases.  For example, this should generate a syntax error:

```
type my_t = my_t
```

amkCha:
Closes #1537 

---

> [!NOTE]
> **Medium Risk**
> Touches core parsing/linking/type-checking and data encoding/decoding paths to introduce alias resolution, which could affect compilation semantics and runtime I/O packing. Cycle handling is noted as incomplete (one cyclic-alias invalid test is disabled), increasing risk of edge-case regressions.
> 
> **Overview**
> Adds first-class **type aliases** to ZkC via `type Name = <type>` declarations, and documents the feature in `ZKC_LANGUAGE.md`.
> 
> Introduces an `Alias` type in the AST (`data.Alias`) and threads alias awareness through the compiler pipeline: lexer/parser recognize `type` declarations and treat unknown type identifiers as alias references; the linker resolves alias symbols (`symbol.TYPE_ALIAS`) into `ResolvedTypeAlias` declarations; and type operations (`BitWidthOf`, `SubtypeOf`, `EquiTypes`, `Flattern`, and encode/decode) now resolve aliases to their underlying types for compatibility checks and I/O packing.
> 
> Extends validation/codegen/program I/O helpers to ignore type-alias declarations as non-executable components, and adds unit/invalid fixtures covering alias use in constants/expressions/casts and expected errors (including cyclic alias cases).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c289d6f72385c1e4d144e68d489e2aac6201ae94. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/zkc/compiler/ast/data/alias.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Flattern implementation for Type interface
+func (p *Alias[I]) Flattern(prefix string, constructor func(name string, bitwidth uint)) {
+	constructor(prefix, p.bitwidth)
+}
+
+// TODO
+func (p *Alias[I]) String() string {
+	return fmt.Sprintf("u%d", p.bitwidth)
```
amkCha: Should it be distinct from a fundamental type ?

pkg/zkc/compiler/ast/data/alias.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// Alias captures the alias of a language type.
+type Alias[I symbol.Symbol[I]] struct {
+	Name     string
+	bitwidth uint
+}
+
+// NewAlias constructs an alias for a given type.
+func NewAlias[I symbol.Symbol[I]](name string, bitwidth uint) *Alias[I] {
```
DavePearce: Having an explicit `bitwidth` argument here is tricky to make work, unfortunately.  The problem is that this value cannot safely be known until after cycle detection has completed (as, otherwise, you risk entering an infinite loop).  Since the suggested place for cycle detection is after linking, then trying to compute this value during linking will fail (see below)

There are a few different ways to try and solve this problem.  Perhaps the obvious option is to modify `Type.BitWidth()` to accept a function like `BitWidth(mapping func(n I) Type)` and likewise for `Flattern`.  This means changing `Type` to `Type[I any]` which is annoying, but not completely unexpected.  This approach will work, provided `Type.BitWidth()` is never called before the cycle detection check.  Since this is called during the `typing` phase, it makes sense that you would want to call cycle detection before typing (but you could leave that for a separate issue anyway).

So, what I don't like about my suggested approach here is that we have this function `BitWidth()` which can only safely be called at a certain point in the program.  Anyway, there are options to address ... but let's not worry about it right now.

pkg/zkc/compiler/ast/data/alias.go
DavePearce: Yes.  I would just return its name here.

pkg/zkc/compiler/linker.go
```diff
@@ -172,6 +187,19 @@ func (p *Linker) linkFunction(fn ast.UnresolvedFunction) (ast.Declaration, []sou
...
 }
 
+func (p *Linker) resolveAlias(d *ast.UnresolvedAlias) *ast.Alias {
+	s, okBus := p.busmap[d.Name]
+	index := s.Index
+	component := p.components[index]
+
+	c, okType := component.(*ast.UnresolvedTypeAlias)
+	if !okBus || !okType {
+		panic("unknown type alias")
```
DavePearce: Yeah, you want to return a syntax error here.

DavePearce: Minimum program to get a `panic`:
```
fn id(x:uint) -> (r:uint) {
   r = x
   return
}

fn uint() -> () {
  return
}
```

pkg/zkc/compiler/linker.go
```diff
@@ -144,22 +145,36 @@ func (p *Linker) linkDeclaration(index uint) (ast.Declaration, []source.SyntaxEr
...
 
 func (p *Linker) linkFunction(fn ast.UnresolvedFunction) (ast.Declaration, []source.SyntaxError) {
 	var (
 		codes = make([]ast.Stmt, len(fn.Code))
 		errs  []source.SyntaxError
 	)
+	// resolve datatype of variables
+	for i, v := range fn.Variables {
+		if d, ok := v.DataType.(*ast.UnresolvedAlias); ok {
+			fn.Variables[i].DataType = p.resolveAlias(d)
```
DavePearce: This is the duplicate code you have which  can also be moved into a `linkType()` function.

DavePearce (commented):
Ok, so lots of comments.  But, overall, you are going in the right direction.  Much of it looks good to me, such as updates to the parser, etc.  Also, its difficult because I am imagining how we are going to use this feature, but that's not exactly easy to predict.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -367,6 +369,37 @@ func (p *Parser) parseReadWriteMemory() (ast.UnresolvedDeclaration, []source.Syn
...
+	if _, errs := p.expect(KEYWORD_TYPE_ALIAS); len(errs) > 0 {
+		return nil, errs
+	} else if name, errs = p.parseIdentifier(); len(errs) > 0 {
+		return nil, errs
+	} else if _, errs = p.expect(EQUALS); len(errs) > 0 {
+		return nil, errs
+	} else if datatype, errs = p.parseType(); len(errs) > 0 {
+		return nil, errs
+	}
+	// check for cyclic definition
```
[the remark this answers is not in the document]
amkCha: I removed it :ok_hand:

pkg/zkc/compiler/ast/data/alias.go
```diff
@@ -0,0 +1,56 @@
...
+
+// NewAlias constructs an alias for a given Type.
+func NewAlias[I symbol.Symbol[I]](name I) *Alias[I] {
+	return &Alias[I]{name}
+}
+
+// AsUint implementation for Type interface
+func (p *Alias[I]) AsUint(Environment[I]) *UnsignedInt[I] {
+	return nil
+}
```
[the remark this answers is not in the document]
amkCha: This gives an `expected uint` error 
```
type bool = u1
const TRUE:bool = 1
const FALSE:bool = ~TRUE
``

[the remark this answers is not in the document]
amkCha: Will distinguish `AsUint` method and `AsAliasOfUint`

PR #1631, opened 2026-04-07
perf: speed up computed register expansion (6.9x trace expansion)

gbotrel:
## Summary

Two optimizations that reduce trace expansion time for the Linea prover from ~2m51s to ~24.6s (6.9x speedup on 96 cores):

1. **Parallel non-recursive fwdComputation** — For non-recursive computed registers, each row's evaluation is independent. Split the row range into chunks (one per CPU core) and process in parallel goroutines. Falls back to sequential for small heights (<4096).

2. **Compiled PolyFil fast path + direct field write** — Pre-compile `PolyFil` expressions once (resolve column references upfront), then evaluate the cached representation per row. Write results directly to field elements instead of going through `word.BigEndian` → `SetBytes` → field round-trip. Added a `Uint64` fast path in `columnAdapter.Get` to avoid expensive `SetBytes` for small values.

## Changes

| File | Change |
|------|--------|
| `pkg/ir/assignment/computed_register.go` | Parallel dispatch for non-recursive, `compiledPolyFil` cached evaluator, `fieldWordSplitter` for direct field writes |
| `pkg/schema/agnostic/filler.go` | Export `Polynomial()` and `RightShift()` on `PolyFil` for compiled evaluation |
| `pkg/trace/util.go` | `Uint64` fast path in `columnAdapter.Get` — avoids `SetBytes` for values that fit in uint64 |

## Benchmark

Measured on r8a.24xlarge (96c AMD EPYC 9R45, 741 GB) with Linea mainnet blocks 29994327-29994333 (22 txs, 7.3M gas):

| Configuration | Trace Expansion Time |
|---|---|
| go-corset v1.2.7 (upstream) | 2m 51s |
| This branch | 24.6s |
| **Speedup** | **6.9x** |

Impact on full Linea prover wall-clock (on-the-fly path): saves ~2.3 minutes (17.4 min → 15.1 min).

## Test plan

- [ ] `go test ./...` passes
- [ ] Trace expansion produces identical outputs (verified via full prover end-to-end test producing valid proof)
- [ ] No regression for recursive computed registers (still use sequential path)

---

> [!NOTE]
> **Medium Risk**
> Introduces new parallel evaluation paths and low-level bit/field conversions during trace expansion, which can surface race/ordering bugs or subtle value-splitting differences despite being performance-focused.
> 
> **Overview**
> Speeds up computed-register trace expansion by adding **parallel, non-recursive forward computation**, splitting rows across goroutines with a small-trace sequential fallback.
> 
> Adds a `PolyFil` fast path that **pre-compiles polynomial fillers** (resolving column accesses once) and evaluates them per-row, plus a `fieldWordSplitter`/`materializeFieldColumns` path to write limb results directly into field elements instead of round-tripping via `word.BigEndian`.
> 
> Exposes `PolyFil.Polynomial()`/`RightShift()` for compilation, and adds a `Uint64` shortcut in `trace.columnAdapter.Get` to avoid expensive byte conversions for small values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa0d483ae86ae321a7794253952c983fa7ca2b6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/ir/assignment/computed_register.go
+++ b/pkg/ir/assignment/computed_register.go
@@ -23,12 +23,14 @@ import (
 	"github.com/consensys/go-corset/pkg/ir/term"
 	"github.com/consensys/go-corset/pkg/schema"
 	sc "github.com/consensys/go-corset/pkg/schema"
+	"github.com/consensys/go-corset/pkg/schema/agnostic"
 	"github.com/consensys/go-corset/pkg/schema/constraint"
 	"github.com/consensys/go-corset/pkg/schema/module"
 	"github.com/consensys/go-corset/pkg/schema/register"
 	"github.com/consensys/go-corset/pkg/trace"
 	"github.com/consensys/go-corset/pkg/util"
 	"github.com/consensys/go-corset/pkg/util/collection/array"
+	"github.com/consensys/go-corset/pkg/util/collection/bit"
 	"github.com/consensys/go-corset/pkg/util/field"
 	"github.com/consensys/go-corset/pkg/util/source/sexp"
 	"github.com/consensys/go-corset/pkg/util/word"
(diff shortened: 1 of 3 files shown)


DavePearce:
Overall, I agree this improves performance and really helps to clarify areas where time is spent.  However, I do not like the hard-coded constants (e.g. `1024` for determining batch size) and also the direct use of `GOMAXPROCS`.  The latter essentially conflicts with parallelisation used in the outermost loop in `ParallelTraceExpansion`.  Ideally, these things would be configurable from the command line (i.e. as the batch size used in `ParallelTraceExpansion` is).

That said, overall, this appears to be a clear performance win for our use cases.

DavePearce:
Subsumed by #1645 I believe.

PR #1643, opened 2026-04-13
1536 feat fixed-size array

amkCha:
> [!NOTE]
> **Medium Risk**
> Medium risk because it extends the language type system and compiler pipeline (parser/linker/typechecker/lowering/codegen) to introduce `FixedArray`, and bugs could surface as mis-typed programs or incorrect lowering into scalar registers.
> 
> **Overview**
> Adds **fixed-size array** support to the ZKC language, including new syntax `[T;N]` (with `N` as a decimal literal or constant) plus `x[i]` read/write expressions.
> 
> Extends the AST `Type` interface with `AsFixedArray`, introduces `data.FixedArray`, updates type equivalence/subtyping/bitwidth, and wires arrays through parsing, linking, and typing (including compile-time constant index requirement, bounds checks, and resolving size constants).
> 
> Introduces a new lowering pass `lower.FlattenFixedArrays` run after validation that expands fixed-array locals/params/returns into scalar variables and rewrites array accesses/whole-array assignments and function arguments accordingly; codegen now assumes array accesses are fully lowered (panics if reached). Formatting is adjusted to avoid inserting spaces in `[u8;3]`. Extensive unit/invalid test cases are added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f04cbf97463d113998e61f988761fa6ddc17ee0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/zkc/compiler/ast/data/bitwidth.go
```diff
@@ -28,6 +28,8 @@ func BitWidthOf[S symbol.Symbol[S]](t Type[S], env Environment[S]) uint {
 		return t.bitwidth
 	case *Alias[S]:
 		return BitWidthOf(t.Resolve(env), env)
+	case *FixedArray[S]:
+		return BitWidthOf(t.Resolve(env), env)
```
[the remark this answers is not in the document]
amkCha: yes, but we compare the underlying bitwidth for typing on this

pkg/zkc/compiler/validate/typing.go
```diff
@@ -202,6 +203,24 @@ func (p *TypeChecker) typeLval(target LVal, env VariableMap, effects bit.Set) (T
...
+			ith_t, errs := p.typeExpression(nil, e, variable.ArrayMap[symbol.Resolved](), effects)
+			errors = append(errors, errs...)
+
+			if len(errs) == 0 && ith_t.AsUint(p.env) == nil {
+				errors = append(errors, *p.srcmaps.SyntaxError(e, "expected uint"))
+			} else if len(errs) == 0 {
+				errors = append(errors, p.checkFixedArrayBounds(e, fixedArr.Size)...)
+			}
+		}
+		return fixedArr.DataType, errors
```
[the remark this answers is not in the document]
amkCha: In this switch case, the target is an lval.Array so the variable is a fixed array

pkg/zkc/compiler/compiler.go
```diff
@@ -70,10 +70,22 @@ func Compile(files ...source.File) (ast.Program, source.Maps[any], []source.Synt
...
 	//
 	errors = append(errors, linkErrs...)
-	// Flatten block-level constructs (if/else, while, for) into flat if-goto form
+	// Error check
+	if len(errors) != 0 {
+		return ast.Program{}, srcmaps, errors
+	}
+	// Lower block-level constructs (if/else, while, for) into flat if-goto form
+	// so that the type checker sees IfGoto statements (not IfElse).
 	lower.Flatten(program, srcmaps)
```
amkCha, line 74: Does this step have to be done prior to validateProgram ? 
I had a lowering step for fixed array done after typing because that's where the compile time constant is checked, but should I modify this lowering step to actually check the compile time constantness so it can run before validateProgram ? I guess what I'm asking is if the two lowering steps need to be adjacent and below to the same phase (as I'm not sure what block level construct is needed for)

amkCha: actually trying it, I'm not sure I could put my lowering phase before validateProgram. Indeed I need to evaluate Args of lval.array which datatypes are set in typing phase (line 468)

pkg/zkc/compiler/ast/expr/array_access.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/consensys/go-corset/pkg/util/collection/set"
+	"github.com/consensys/go-corset/pkg/zkc/compiler/ast/data"
+	"github.com/consensys/go-corset/pkg/zkc/compiler/ast/symbol"
+	"github.com/consensys/go-corset/pkg/zkc/compiler/ast/variable"
+)
+
+// ArrayAccess represents an array access within an expression.
+type ArrayAccess[S symbol.Symbol[S]] struct {
+	Id       variable.Id
+	Args     []Expr[S]
```
DavePearce: This looks odd to me.  You only expect a single index in an array access, so you should reflect that here (i.e. `Arg Expr[S]` not `Args []Expr[S]`.  I investigated what happens when you provide multiple indices, and it leads to a `panic`.  Here's my test:

```
fn getElement(i:u2, x:[u32;4]) -> (r:u32) {
  if i == 0 {
    r = x[0,0]
  } else {
    r = x[1]
  }
}
```

DavePearce: In the parser you do this:

```
// Array access
var args []Expr
//
args, errs = p.parseExprList(RSQUARE, env)
//
nexpr = expr.NewArrayAccess(rid, args...)
```

Here, you could just do this instead:
```
var arg Expr
//
if arg, errs = p.parseExpr(env); len(errs) > 0 {
	return nil, errs
} else if _, errs = p.expect(RSQUARE); len(errs) > 0 {
	return nil, errs
}
//
nexpr = expr.NewArrayAccess(rid, arg)
```

pkg/zkc/compiler/ast/lval/array.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/consensys/go-corset/pkg/zkc/compiler/ast/symbol"
+	"github.com/consensys/go-corset/pkg/zkc/compiler/ast/variable"
+)
+
+// Array represents an array write within an assignment.
+type Array[S symbol.Symbol[S]] struct {
+	// name of the array being written
+	Id variable.Id
+	// identifies location being written
+	Args []expr.Expr[S]
```
DavePearce: As above for `ArrayAccess` we want `Arg expr.Expr[S]` here.

DavePearce: Similar test to before as well:

```
fn getElement(i:u2, x:u32) -> (r:[u32;4]) {
  r[0,0] = x
}
```

pkg/zkc/compiler/format/insert_rules.go
```diff
@@ -78,7 +78,7 @@ func init() {
 	DEFAULT_INSERTION_RULES[parser.BITWISE_SHR] = InsertSpaceAround()
 	// Space after separators.
 	DEFAULT_INSERTION_RULES[parser.COMMA] = InsertSpaceAfter()
-	DEFAULT_INSERTION_RULES[parser.SEMICOLON] = InsertSpaceAfter()
+	DEFAULT_INSERTION_RULES[parser.SEMICOLON] = InsertSemicolonSpace()
```
DavePearce, line 81: Well done --- good job spotting this!

DavePearce: We have to consider the formatting now when adding new syntax :)

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -35,6 +49,335 @@ func Flatten(program ast.Program, srcmaps source.Maps[any]) {
...
+// fixed-size array variables into individual scalar variables.  A variable
+// arrayName of type uM[n] is replaced by n scalars arrayName$0 .. arrayName$(n-1),
+// each of type uM.  Corresponding expr.ArrayAccess and lval.Array nodes are
+// rewritten to plain LocalAccess / lval.Variable references.
+func (p Lowering) FlattenFixedArrays(program ast.Program, srcmaps source.Maps[any]) {
+	env := program.Environment()
+
+	for _, d := range program.Components() {
+		if fn, ok := d.(*decl.ResolvedFunction); ok {
+			if p.expandFixedArrays {
```
DavePearce: tbh, after thinking about it last night, I'm not completely sure of the value of having the `expandFixedArrays` flag here.  Is the use case for debugging?  Basically so we can see how it looks without expansion being applied.  If so, then I guess that seems OK.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -35,6 +49,335 @@ func Flatten(program ast.Program, srcmaps source.Maps[any]) {
...
+	case *expr.ArrayAccess[symbol.Resolved]:
+		rewriteFixedArrayExprs(e.Args, mapping, declarations, env)
+
+		m := mapping[e.Id]
+		if !m.isArray {
+			e.Id = m.newBase
+			return e
+		}
+
+		if len(e.Args) != 1 {
```
DavePearce: This will go once you change `Args` to `Arg.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -35,6 +49,335 @@ func Flatten(program ast.Program, srcmaps source.Maps[any]) {
...
 
+// FlattenFixedArrays walks every component in the program and expands
+// fixed-size array variables into individual scalar variables.  A variable
+// arrayName of type uM[n] is replaced by n scalars arrayName$0 .. arrayName$(n-1),
+// each of type uM.  Corresponding expr.ArrayAccess and lval.Array nodes are
+// rewritten to plain LocalAccess / lval.Variable references.
+func (p Lowering) FlattenFixedArrays(program ast.Program, srcmaps source.Maps[any]) {
+	env := program.Environment()
+
+	for _, d := range program.Components() {
```
DavePearce, line 41: This loop could be made more efficient.  I guess Claude wrote it, but its not that flash.  The issue is that it does a lot of repeated work.  What you want is to have the `varMapping` initialised in this function and then have two loops: the first goes through expanding the function signatures and variables, whilst updating the mapping.  The second then goes through and rewrites all `ArrayAccess` expressions, `Array` lvals and `ExternAccess` expressions.  

Anyway, its not a big drama.  But, it initially confused me compared with what I was expecting.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -35,6 +49,335 @@ func Flatten(program ast.Program, srcmaps source.Maps[any]) {
...
+	default:
+		panic(fmt.Sprintf("unknown statement encountered during fixed-array lowering: %T", s))
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Expression rewriting
+// ---------------------------------------------------------------------------
+
+func rewriteFixedArrayExpr(
```
DavePearce, line 359: Btw, the use of rewriting here looks OK to me.  I was expecting this.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -669,14 +669,76 @@ func (p *Parser) parseType() (Type, []source.SyntaxError) {
...
+		p.srcmap.Put(lookahead, p.spanOf(p.index, p.index))
+
+		if _, errs := p.expect(SEMICOLON); len(errs) != 0 {
+			return nil, p.srcmap.SyntaxErrors(lookahead, "expected semicolon to define array size")
+		}
+
+		lookahead = p.lookahead()
+		p.srcmap.Put(lookahead, p.spanOf(p.index, p.index))
+
+		switch lookahead.Kind {
```
DavePearce: Hmmm, its not that pretty this section of code. I'd suggest making a subfunction which returns `size, sizeName, []source.SyntaxError` or something.

pkg/zkc/compiler/linker.go
```diff
@@ -497,6 +506,33 @@ func (p *Linker) linkType(datatype data.UnresolvedType) (data.ResolvedType, []so
...
+		if t.SizeName != "" {
+			resolved, errs := p.resolveConstantSize(t.SizeName, datatype)
+			if len(errs) != 0 {
+				return nil, errs
+			}
+
+			size = resolved
+		}
+
+		switch d := t.DataType.(type) {
```
DavePearce: I didn't understand this.  Why not call `linkType` recursively here on `t.DataType` ?

pkg/zkc/compiler/linker.go
```diff
@@ -509,6 +545,45 @@ func (p *Linker) linkType(datatype data.UnresolvedType) (data.ResolvedType, []so
 	default:
 		return nil, p.srcmap.SyntaxErrors(datatype, "unknown type encountered")
 	}
+
+	return nil, p.srcmap.SyntaxErrors(datatype, "unknown type encountered")
+}
+
+// resolveConstantSize resolves constant value for use as a
+// fixed-array size.
+func (p *Linker) resolveConstantSize(name string, node any) (uint, []source.SyntaxError) {
```
DavePearce: Overall, this is somewhat problematic.  You're trying here to do what the later function `EvalConstant` does --- but this is not a good place to do it.  Examples like this cause problems:

```
const SIZE:u8 = 2
const SIZE_P1:u8=SIZE+1

fn fg(i:u2, x:[u32;SIZE_P1]) {

}
```

If it was me, I would avoid this rabbit hole (at least for the first PR) and simply require that the size is always given as an explicit numeric literal.  Anyway, up to you --- its not a big deal.

pkg/zkc/compiler/compiler.go
```diff
@@ -70,10 +70,22 @@ func Compile(files ...source.File) (ast.Program, source.Maps[any], []source.Synt
 	program, srcmaps, linkErrs = Link(items...)
 	//
 	errors = append(errors, linkErrs...)
-	// Flatten block-level constructs (if/else, while, for) into flat if-goto form
+	// Error check
+	if len(errors) != 0 {
```
DavePearce: Ok, unfortunately, this is a problem.  I guess its a "new thing" caused by the IDE.  We want to always call `validateProgram()` --- even if there are errors.  This is because we want to report as many errors as possible in the IDE.  You have to make your code "resilient" to this.  It means what cursor says above about typing the `lval` is correct --- you need to handle the case where the `DataType` is nil.  Something like this is the example:

```
//type my_t = u32

fn fg(i:u2) -> (r:my_t) {
   r[0] = 1
}
```

This is tricky and I spent quite a lot of time making my existing typing code work well enough for the IDE.  But, even I cut a few corners which you can see in `validateProgram` where it still terminates early if it finds a cycle.  I intend to fix that at some point, but it requires completely reworking the way cycle detection works.

pkg/zkc/compiler/compiler.go
```diff
@@ -70,10 +70,22 @@ func Compile(files ...source.File) (ast.Program, source.Maps[any], []source.Synt
...
+		return ast.Program{}, srcmaps, errors
+	}
+	// Lower block-level constructs (if/else, while, for) into flat if-goto form
+	// so that the type checker sees IfGoto statements (not IfElse).
 	lower.Flatten(program, srcmaps)
 	// Well-formedness checks (assuming unlimited field width).
 	errors = append(errors, validateProgram(program, srcmaps)...)
+	// Error check
+	if len(errors) != 0 {
+		return ast.Program{}, srcmaps, errors
```
DavePearce: Also a problem here (similar to the error check above).  We don't want to return `ast.Program{}` but instead to return what we have.  In this case, I think its OK to not call `FlattenFixedArrays` when there are errors as it doesn't itself generate any syntax errors.

pkg/zkc/compiler/validate/typing.go
```diff
@@ -594,6 +616,31 @@ func (p *TypeChecker) typeLocalAccess(e *expr.LocalAccess[symbol.Resolved], env
...
 
+func (p *TypeChecker) typeArrayAccess(e *expr.ArrayAccess[symbol.Resolved], env VariableMap,
+) (Type, []source.SyntaxError) {
+	var (
+		errors  []source.SyntaxError
+		effects bit.Set
+	)
+
+	varType := env.Variable(e.Id).DataType
+	fixedArr := varType.AsFixedArray(p.env)
```
DavePearce: CursorBot is right that its a problem here to call `varType` directly because it could be `nil`.  The `wellFormed()` function is intended to be used to check whether or not a `data.Type` is valid or not.  If its not valid, then you have to be careful about which methods you call.  I think for example, calling `t.AsUint()` when `!wellFormed(t, p.env)` holds can lead to crashes.

It can be hard to figure out the best way to write the code in a "resilient" style.  It will be easier once you have only `Arg` and not `Args`.  I would do this: (i) run the `ith_t.AsUint()` check only when `wellFormed(ith_t, p.env)` holds (it doesn't matter if `len(errors) > 0` though); run the `p.checkFixedArrayBounds(e, fixedArr.Size)` check only when `wellFormed(ith_t, p.env)` and `wellFormed(varType, p.env)` holds (again doesn't matter if `len(errors) > 0` only that both types are well formed).

DavePearce: And I guess a similar situation arises for typing `ArrayAccess`.  Yeah, its a bit annoying doing this, but it does make the IDE experience better.  If you struggle to get this working, you could also just leave it for me.

pkg/zkc/compiler/compiler.go
DavePearce: Yeah, thinking about this.  I think the issue is that you want `validateProgram` to run first so that we know if there were any errors (i.e. to not do your lowering step if there are).  That makes sense to me.

In principle, `lower.Flattern()` could run after `validateProgram` and then we can combine the two lowerings together.  But its tricky as we have to update the various steps used in `validateProgram`.  I think `Typing()` is easy to update, but `ControlFlow` is not.  

Overall, I think for now we leave as is. At some point maybe we can do something here, but at the same time ... it works as is :)

DavePearce (changes requested):
Ok, overall, that's quite a few comments in the end --- sorry.  There are a surprising amount of details, even for a feature like this.

And, I still have two more problematic examples.  I think maybe we should address these in separate issues:

```
fn fg(i:u32) -> (r:[u32;2]) {
   r[0] = i
   return
}
```
This program should report an error, but currently doesn't.  Its not easy to fix this I guess.

```
fn fg(x:[u32;2]) -> (r:[u32;2]) {
   r = x
   return
}
```

This program causes a compiler `panic`.  Personally, I think this case is doable and we would want the same for `struct`s.  But its fine to do it in a separate issue.

The IDE presents a key problem for a compiler `panic`.  Specifically, you don't want the IDE to crash on you when you are typing code, even if you know that code is wrong or something.

testdata/zkc/invalid/type_10.zkc
```diff
@@ -1,8 +1,7 @@
-//error:10:20-21:unknown type alias
-//error:6:8-9:unknown type alias
-//error:6:17-18:unknown type alias
-//error:7:11-12:unknown type alias
-//error:12:2-3:return y possibly undefined
```
DavePearce, line 5: Yeah, so here is one test which is now failing because `Compile()` returns early when there are some errors and, hence, we lose an error (which you've removed here).  What this tells me is that I should add more tests which cross the various boundaries.  Specifically, tests where we have a linking failure and a typing failure, etc.

pkg/zkc/compiler/ast/expr/array_access.go
amkCha: :100: addressed comment and added an invalid test

```
//error:7:19-20:unexpected token
pub output res(address:u8) -> (byte:u8)
fn main() {
  var items:[u8;2]
  items[0] = 1
  items[1] = 1
  res[0] = items[0,0]
}
```

pkg/zkc/compiler/ast/lval/array.go
amkCha: Added a new test
```
fn main() {
  var items:[u8;2]
  items[0,0] = 8
}
```

pkg/zkc/compiler/lower/flatten.go
amkCha: agreed :+1:

pkg/zkc/compiler/format/insert_rules.go
amkCha: another detail that came along :laughing:

testdata/zkc/invalid/type_10.zkc
amkCha: it's back :)

pkg/zkc/compiler/linker.go
amkCha: :100: I can't remember why I thought I couldn't

pkg/zkc/compiler/parser/parser.go
amkCha: true, revamped a bit with a `parseArraySize` :+1:

pkg/zkc/compiler/linker.go
amkCha: Ok good point - in the end, I removed the "attempt" to resolve anything here and moved it in the check for array bounds in the typing phase. Easier to do it as the constants are resolved already and also, it looked to me it made more sense

amkCha: This test is now passing 
```
const SIZE:u4 = 3
const SIZE_P1:u4=SIZE+1
fn main() {
    var items:[u8;SIZE_P1]
    items[2] = 0
}
```

pkg/zkc/compiler/compiler.go
amkCha: I added your test, that now bubbles up several errors 
```
//error:4:12-16:unknown type alias
//error:5:5-9:fixed-array has unresolved type
 fn main() -> () {
    var r:[my_t;2]
    r[0] = 1
 }
```

pkg/zkc/compiler/validate/typing.go
amkCha: Got it, I ended up writing `typeArray` that can be used for both `lval` and `array_access` so that the logic is in one place and used `wellFormed`

amkCha:
> Ok, overall, that's quite a few comments in the end --- sorry. There are a surprising amount of details, even for a feature like this.
> 
> And, I still have two more problematic examples. I think maybe we should address these in separate issues:
> 
> ```
> fn fg(i:u32) -> (r:[u32;2]) {
>    r[0] = i
>    return
> }
> ```
> 
> This program should report an error, but currently doesn't. Its not easy to fix this I guess.
> 
> ```
> fn fg(x:[u32;2]) -> (r:[u32;2]) {
>    r = x
>    return
> }
> ```
> 
> This program causes a compiler `panic`. Personally, I think this case is doable and we would want the same for `struct`s. But its fine to do it in a separate issue.
> 
> The IDE presents a key problem for a compiler `panic`. Specifically, you don't want the IDE to crash on you when you are typing code, even if you know that code is wrong or something.

With the help of Claude, I changed the lowering in `flatten.go` to accept and not panic on the 2nd example. The 2st program I didn't try because I wasn't sure why we should report an error ? Is it because the array is not fully defined ?

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -13,16 +13,407 @@
...
+			Datatype: src.Type(),
+		}
+
+		target := lval.NewArray[symbol.Resolved](lv.Ids[0], expr.NewConstant[symbol.Resolved](idx, 10))
+
+		result[i] = &stmt.Assign[symbol.Resolved]{
+			Targets: []lval.LVal[symbol.Resolved]{target},
+			Source:  access,
+		}
+	}
```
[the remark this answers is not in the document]
amkCha: Hmm, the test Test_ZkcUnit_FixedArray_08 does exactly this r = x whole-array assignment path, so I confirm it works correctly.

pkg/zkc/compiler/lower/flatten.go
amkCha: Okay the loop is now refactored. A bit more tricky for the expansion as so were nested but I guess it does make it cleaner/clearer

pkg/zkc/compiler/validate/typing.go
```diff
@@ -594,6 +599,37 @@ func (p *TypeChecker) typeLocalAccess(e *expr.LocalAccess[symbol.Resolved], env
...
+
+	// Check argument is unsigned integers and well-formed
+	arg_t, errs := p.typeExpression(nil, arg, env, effects)
+	arg_t_ok := wellFormed(arg_t, p.env)
+
+	errors = append(errors, errs...)
+
+	//
+	if !varType_ok {
+		errors = append(errors, *p.srcmaps.SyntaxError(arg, "fixed-sized array has unresolved type"))
```
DavePearce: Ok, so you don't need to report an error when the type is not well-formed.  If its not well-formed then you can assume that an error has already been reported somewhere up the chain (hence you're duplicating the error here in some sense).  In this case, you can just return `nil` with whatever errors have been generated.

pkg/zkc/compiler/validate/typing.go
```diff
@@ -702,6 +738,45 @@ func (p *TypeChecker) checkCastType(to, from Type, node any) []source.SyntaxErro
...
+	if ko != "" {
+		return p.srcmaps.SyntaxErrors(arg, "array index must be a constant expression")
+	}
+	// Resolve Size from SizeName
+	resolved := false
+
+	if fixedArray.SizeName != "" {
+		for _, d := range p.program.Components() {
+			if c, ok := d.(*decl.ResolvedConstant); ok && c.Name() == fixedArray.SizeName {
+				valSize, ko := codegen.EvalConstant(c.ConstExpr, false, p.program.Components(), p.env)
```
DavePearce: `ko` here provides a suitable error message to use which tells you more about what went wrong.  I feel like its better to use that error message rather than create a blanket message which hides the real reason there was a problem.

DavePearce (approved):
Ok, I think it looks good.  I left two comments, but they are not super important.  Ready to merge!

PR #1680, opened 2026-04-23
Switch statement support in zkc

### LINKED ISSUE #1623: feat: support `switch` statement in ZkC
# Summary

There is a need to support `switch` statements in ZkC.   This should differ slightly in semantics (i.e. meaning) from `switch` statements in e.g. C.  **Specifically, it should not support automatic fall thru semantics**.  Instead, it should follow `go` semantics for `switch` statements (see [here](https://go.dev/tour/flowcontrol/9)).

## Example

Here is a proposed example syntax:

```
pub input data(address:u16) -> (byte:u8)

fn main() {
  var val:u8 = data[0]
  var res:u8 = data[1]

  if example(val) != res {
    fail
  }
}

fn example2(val:u8) -> (r:u8) {
   switch val {
   case 0:
     r = 0
   case 1:
     r = 1
   case 2,3:
     r = 1
   default:
     r = 2
   }
   
   return
}
```

Note here that `case 2,3:` illustrates the ability to specify multiple cases together.  Furthermore, `example` is equivalent to the following:

```
fn example(val:u8) -> (r:u8) {
   if val == 0 || val == 1 {
      r = 0
   } else if val == 2 {
      r = 1
   } else if val == 3 {
      r = 1
   } else {
      r = 2
   }
   //
   return
}
```

Some test cases would be:

```
{ "data": "0x0000" }
```
```
{ "data": "0x0301" }
```

OlivierBBB:
> [!NOTE]
> **Medium Risk**
> Adds a new control-flow construct that is parsed, linked, and lowered into `if`/`goto`, which can affect control-flow validation and codegen behavior. Risk is mitigated by extensive new unit/invalid test coverage, but edge cases (e.g., label duplication) appear not fully enforced yet.
> 
> **Overview**
> Adds `switch` statement support end-to-end: the lexer/parser now recognize `switch`/`case`/`default`, build a new `stmt.Switch` AST node, and the linker resolves its discriminant/labels/bodies.
> 
> The lowering pass now converts `switch` into nested `if/else` chains (moving any `default` to the final `else`) so downstream passes still operate on flat `if-goto` form; formatter and LSP semantic token mapping were updated to treat the new keywords correctly.
> 
> Introduces new `zkc` unit/invalid test fixtures for `switch` behavior and errors (e.g., multiple defaults, invalid labels, incomplete returns), plus a small comment-only rename in `expr.Expr` (`RegistersRead` -> `LocalUses`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8dbf8bc0d64dba1013a13d79dc2679c6ad6e393e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


testdata/zkc/invalid/switch_duplicate_label_01.zkc
```diff
@@ -0,0 +1,19 @@
+//error:17:18-23:duplicate label in switch statement
```
DavePearce, line 1: @OlivierBBB These tests need to be added to `zkc_invalid_test.go` in order to be run.  At the moment, none of your invalid tests are being run.  In this particular case (duplicate labels), if you didn't implement the check yet, you might leave the test commented out in `zkc_invalid_test.go` and raise an issue to fix that later.

testdata/zkc/invalid/switch_incomplete_return.zkc
```diff
@@ -0,0 +1,34 @@
+//error:33:3-8:return y possibly undefined
```
DavePearce: Yeah, this is a good test.

testdata/zkc/unit/switch_notes.md
```diff
@@ -0,0 +1 @@
+- switch: type consistency between discriminant and cases (numeric or constants)
```
DavePearce: Not sure this is helpful.  We generally don't leave notes in the tests directory --- its not consistent.

testdata/zkc/unit/switch_04.accepts
```diff
@@ -0,0 +1,4 @@
+{ "input_case": "0x00", "expected_output": "0xff" }
+{ "input_case": "0x01", "expected_output": "0x04" }
+{ "input_case": "0x02", "expected_output": "0x09" }
+{ "input_case": "0x03", "expected_output": "0x11" }
```
DavePearce: It would not hurt to add a few more values for the `default` case.  Chose some random `u8`s maybe

testdata/zkc/unit/switch_06.zkc
```diff
@@ -0,0 +1,37 @@
+// =============================================
```
DavePearce, line 1: Would be good to have some `accepts` inputs for this test!

testdata/zkc/unit/switch_04.zkc
```diff
@@ -0,0 +1,36 @@
+const CASE_A:u8 = 1
```
DavePearce, line 1: I think you're missing a test case involving e.g. `case 0, 1: ...`

DavePearce: Also, I'd suggest a test of this form:
```
   var y:u32

   switch x {
      case 0: {
        r = 0
        return
      }
      case 1: {
        y = 0
      }
      default: {
        y = x + 1
      }
   }
   //
   r = y
```

This is checking that the tool notices that `y` does not need to be assigned in the first `case`, and concludes that `y` is definitely assigned by the statement `r = y`.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -825,6 +827,207 @@ func (p *Parser) parseCallStatement(env Environment) (stmt.Unresolved, []source.
...
+	p.expect(RCURLY)
+
+	// for a switch statement to 'return' as a whole, and in the absence of
+	// exhaustivity analysis, it must hold that
+	//	- every branch 'returns'
+	//	- its body contains a default branch which similarly 'returns'
+	//
+	// Note. The required presence of a default branch prevents switch
+	// statements with "empty" body from 'returning'.
+	returned = bodyContainsDefaultBranch && everyBranchReturns
```
DavePearce, line 990: The logic you have written here looks good to me.  But, this test is not parsing:
```
fn f(x:u32) -> (r:u32) {
    switch x {
        default: {
            r = x
            return
        }
    }
}
```

I'm not sure exactly why (could be something to do with lowering I think, whereby unnecessary `goto` instructions are added).  This is an edge case though.  Potentially just raise an issue with this as an example to look into.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -825,6 +827,207 @@ func (p *Parser) parseCallStatement(env Environment) (stmt.Unresolved, []source.
...
+	for _, expression := range cases {
+		switch t := expression.(type) {
+		case *expr.ExternAccess[symbol.Unresolved]:
+			{
+				switch t.Name.Kind {
+				// the 'named constant' case
+				case symbol.CONSTANT:
+					continue
+				default:
+					return nil, nil, p.syntaxErrors(p.previousToken(), "switch case options should be constants, named or litteral")
```
DavePearce: I agree with CursorBot that using `p.previousToken()` doesn't make sense here.  Instead, you have the source mapping infoprmation available on `expression` so you can just use that.  So something like this will do it:

```
return nil, nil, p.srcmap.SyntaxErrors(expression, "expected constant or literal")
```

DavePearce (commented):
Overall, its almost there --- I've left a few comments.  Also, you need to add rules for the formatter as well.  Specifically, in `insert_rules.go` you need the following lines:
```
DEFAULT_INSERTION_RULES[parser.KEYWORD_SWITCH] = InsertAfter(ONE_SPACE)
DEFAULT_INSERTION_RULES[parser.KEYWORD_CASE] = InsertAfter(ONE_SPACE)
```

Without these lines, you'll notice that the formatter turns e.g. `switch x` into `switchx` which is invalid.  I have VS code setup to run the formatter whenever I save, and that works really well --- hence how I noticed this.

testdata/zkc/invalid/switch_duplicate_label_01.zkc
OlivierBBB: Done. The duplicate label tests are commented out + there is an issue on them #1695

pkg/zkc/compiler/parser/parser.go
OlivierBBB: Indeed. Fixed

testdata/zkc/unit/switch_04.zkc
OlivierBBB: > I think you're missing a test case involving e.g. `case 0, 1: ...`

I have a test `switch_numeric_01.zkc` (which may not have been included in the zkc unit test file) that has such statements e.g.
```
  switch discr {
    case 13, 33: { fail }
    default: {}
  }

  switch discr {
    default: { fail}
    case 13: { fail }
    case 65535: { }
  }
```

Is that what you meant ?

testdata/zkc/unit/switch_04.accepts
OlivierBBB: Done I believe

testdata/zkc/unit/switch_06.zkc
OlivierBBB: Done I believe

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -235,6 +240,105 @@ func lowerTernaryCondition(
...
+			mostNestedIfThenElse = equivalentIfThenElse
+		} else {
+			falseBranch = &stmt.IfElse[symbol.Resolved]{
+				Cond:        &logicalOrOfCases,
+				TrueBranch:  branch.Body,
+				FalseBranch: []stmt.Stmt[symbol.Resolved]{},
+			}
+			mostNestedIfThenElse.FalseBranch = append(mostNestedIfThenElse.FalseBranch, falseBranch)
+			mostNestedIfThenElse = falseBranch
+		}
```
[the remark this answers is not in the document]
OlivierBBB: Addressed in latest commit authored by Claude

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -235,6 +240,106 @@ func lowerTernaryCondition(
...
+		falseBranch          *stmt.IfElse[symbol.Resolved]
+	)
+
+	for _, branch := range s.Branches {
+		// we store the default statement if and when we come across it
+		// and continue to the next branch
+		if branch.IsDefault {
+			defaultStatement = &branch.Body
+			continue
+		}
```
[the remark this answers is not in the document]
OlivierBBB: From what claude tells me this comment does not apply with Go 1.22 and later since 'each iteration creates a fresh loop variable'.

pkg/zkc/compiler/parser/parser.go
OlivierBBB: ~Indeed, I get an `unknown variable` error~ I don't get an error, zkc was not built properly.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -235,6 +240,106 @@ func lowerTernaryCondition(
...
+		if equivalentIfThenElse == nil {
+			equivalentIfThenElse = &stmt.IfElse[symbol.Resolved]{
+				Cond:        &logicalOrOfCases,
+				TrueBranch:  branch.Body,
+				FalseBranch: []stmt.Stmt[symbol.Resolved]{},
+			}
+			mostNestedIfThenElse = equivalentIfThenElse
+		} else {
+			falseBranch = &stmt.IfElse[symbol.Resolved]{
+				Cond:        &logicalOrOfCases,
```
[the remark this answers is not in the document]
OlivierBBB: This error, too, is not an issue as of Go version 1.22, as per Claude. AFAICT this is documented [in this paragraph](https://go.dev/blog/loopvar-preview#the-fix)

> The Fix[¶](https://go.dev/blog/loopvar-preview#the-fix)
> 
> For Go 1.22, we plan to change for loops to make these variables have per-iteration scope instead of per-loop scope.

pkg/zkc/compiler/parser/parser.go
OlivierBBB: Fixed. The issue was with returning always `false` as the `returns` bit in  `parseSwitchCase / parseSwitchDefault`.

OlivierBBB: The relevant tests are `switch_07` and `switch_08`.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -235,6 +240,106 @@ func lowerTernaryCondition(
...
+	for _, branch := range s.Branches {
+		// we store the default statement if and when we come across it
+		// and continue to the next branch
+		if branch.IsDefault {
+			defaultStatement = &branch.Body
+			continue
+		}
+
+		logicalOrOfCases := branch.LogicalOrOfCases(s.Discriminant)
+
```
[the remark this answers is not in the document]
DavePearce: @OlivierBBB What co-pilot says here is true --- is an interesting question whether we want to do anything about it.  I think it could be raised as a separate issue, and we can think about what the right way forward is.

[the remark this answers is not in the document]
OlivierBBB: True, I did see that comment but forgot to address it. Issue https://github.com/Consensys/go-corset/issues/1716

pkg/zkc/compiler/parser/parser.go
```diff
@@ -894,6 +896,218 @@ func (p *Parser) parseCallStatement(env Environment) (stmt.Unresolved, []source.
...
+			{
+				continue
+			}
+		default:
+			return false, nil, nil, p.syntaxErrors(p.previousToken(),
+				"labels in a switch statement must be constants (named or literal)")
+		}
+	}
+
+	// TODO: I have no idea whether the boolean parameter is the right one
```
DavePearce: Need to resolve this.  I can take a look.

DavePearce: I think this actually looks OK in the end.  I didn't find any problematic test cases.  I think you can remove the `TODO` here.

testdata/zkc/unit/switch_04.zkc
DavePearce: > Is that what you meant ?

Yes and no.  The problem is you are not executing that test.  So its like only a half test.  You need an `accept` file for it.

Also, it would be helpful to name it `switch_XX.zkc` so it is consistent with the others.

testdata/zkc/unit/switch_06.zkc
```diff
@@ -0,0 +1,51 @@
...
+    r = results[0]
+
+    var y:u8
+    y = do_something(x)
+
+    if y != r {
+        printf "x = %d\n", x
+        printf "y = %d\n", y
+        printf "r = %d\n", r
+        printf "fuck ...\n"
```
DavePearce: This is not appropriate.

testdata/zkc/unit/switch_07.zkc
```diff
@@ -0,0 +1,26 @@
...
+
+fn choose(b:u1, x:u32, y:u32) -> (r:u32) {
+    if b == 1 {
+        r = x
+    } else {
+        r = y
+    }
+}
+
+fn identity(x:u32) -> (r:u32) {
```
DavePearce: You should add some `accept` tests for this.

testdata/zkc/unit/switch_09.zkc
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    // no match
+    switch discr {
+        case 13, 33: {
+            fail
+        }
+    }
+    switch discr {
+        case 13, 33: {
```
DavePearce: The problem with this test is that you can't really test this piece.  You need a standalone test which has a `case` with multiple labels that is executed.

pkg/test/zkc_invalid_test.go
```diff
@@ -197,6 +197,54 @@ func Test_ZkcInvalid_If_04(t *testing.T) {
 	checkZkcInvalid(t, "zkc/invalid/if_04")
 }
 
+// ===================================================================
+// Switch tests
+// ===================================================================
+
+// func Test_ZkcInvalid_Switch_Duplicate_Label_01(t *testing.T) {
```
DavePearce, line 228: Ok, is fine these are commented out for now.  Maybe create an issue to address them in the future which highlights that you already wrote some tests.

DavePearce (changes requested):
Ok, overall I think you are basically there.  The only issue is that you are still missing some tests.  For example, there is no execution test for a case with multiple labels; likewise, there is no execution test for a switch with no default.  I suggested adding `accept` files for some valid tests you already wrote.

Also, use of certain language in a test (`switch_06`) is not OK.  You need to remove that.

testdata/zkc/unit/switch_09.zkc
DavePearce: Also, change the name to `switch_XX.zkc` please so its consistent with the others!

testdata/zkc/unit/switch_06.zkc
OlivierBBB: Oh geez ^^ yep, my bad

pkg/test/zkc_invalid_test.go
OlivierBBB: I have an issue for that: https://github.com/Consensys/go-corset/issues/1695

pkg/zkc/compiler/parser/parser.go
OlivierBBB: The boolean parameter 'returns' I believe I fixed in the mean time, the comment is outdated.

testdata/zkc/unit/switch_07.zkc
OlivierBBB: It's not clear to me: when do you need accepts / rejects files ?

DavePearce (approved):
LGTM

PR #1727, opened 2026-05-05
feat: lower zkc-native bit operation

### LINKED ISSUE #1709: feat: lowering bitwise operations in ZkC
## Summary

An important part of the constraint generation process is to lower certain "high-level" constructs into a more "low-level" representation.  In this case, we are considering the bitwise operations: `&` `|` `^` `~` `<<` `>>`.  These operations cannot be translated directly into constraints inline.  Instead, they require functions / static reference tables be instantiated to handle their implementation (typically via decomposition of some kind).

## Example

To handle bitwise operations in `zkasm` (the predecessor) to ZkC, we used several functions including the `bit_xoan` family of functions.  Here is a snippet of that:

```
fn bit_xoan_u8(INST u2, ARG_1 u8, ARG_2 u8) -> (RES u8) {
  var arg1_lo, arg1_hi, arg2_lo, arg2_hi, res_hi, res_lo u4
  arg1_hi, arg1_lo = ARG_1
  arg2_hi, arg2_lo = ARG_2
  res_hi = bit_xoan_u4(INST, arg1_hi, arg2_hi)
  res_lo = bit_xoan_u4(INST, arg1_lo, arg2_lo)
  RES = ( res_hi * 2^4 ) + res_lo
  return
}

fn bit_xoan_u4(INST u2, ARG_1 u4, ARG_2 u4) -> (RES u4) {
  var arg1_lo, arg1_hi, arg2_lo, arg2_hi, res_hi, res_lo u2
  arg1_hi, arg1_lo = ARG_1
  arg2_hi, arg2_lo = ARG_2
  res_hi = bit_xoan_u2(INST, arg1_hi, arg2_hi)
  res_lo = bit_xoan_u2(INST, arg1_lo, arg2_lo)
  RES =  ( res_hi * 2^2 ) + res_lo
  return
}

...
```

Here, `xoan` stands for `Xor` `Or` `And` `Not` and we see that `bit_xoan_u8` decomposes into `bit_xoan_u4`, etc.

## Approach

This should be implemented as a function (e.g. `LowerBitwise()`) which transforms a ZkC machine into another ZkC machine.  It should operate on a `vm` machine by removing occurrences of the relevant `vm` instructions.  That is, `LowerBitwise` goes through all of the functions in the machine looking for situations where a bitwise operator is used and, instead, these should be replaced with a function call.  Here is a minimal example to illustrate:

```
fn (x:u32) -> (y:u32) {
  y = x >> 1
}
```

This would then be compiled into something (very roughly speaking) which resembles this:

```
fn (x:u32) -> (y:u32) {
   y = $bitshr_u32(x,1)
}

fn $bitshr_u32(l:u32, r:u32) -> (v:u32) {
   ...
}
```

The example implementation of the function is left open and, most likely, would use a combination of decomposition into smaller instances and/or use of static reference table(s).

The `go` signature of the `LowerBitwise()` function might be something like this:

```
func LowerBitwise[W word.Word[W]](vm machine.Base[W]) -> machine.Base[W] {
  ...
}
```

Observe that this means it operates over the `vm` instructions, not the abstract syntax tree.  The reason for this is primarily that we want the ability to execute a `vm` machine which still has bitwise operations (i.e. for performance).

letypequividelespoubelles:
> [!NOTE]
> **Medium Risk**
> Adds a new optional codegen pass that rewrites VM bitwise/shift instructions into generated helper modules and arithmetic ops, which can impact program semantics/performance and module ID layout. Risk is moderated by being opt-in via config/CLI flag and expanded unit coverage running both modes.
> 
> **Overview**
> Adds an opt-in `LowerZkcNative` codegen pass to lower ZkC “native” bitwise/shift VM instructions into arithmetic instructions via generated helper `CALL` modules, running before vectorization.
> 
> The lowering pipeline includes a pre-pass to binarize n-ary `AND/OR/XOR`, special-cases `NOT` as inline `(2^w-1) - x`, and implements `SHL/SHR` with self-recursive helpers and shift-amount width normalization; new unit tests cover casts, chained bitwise ops, and shifts, and existing bitwise/shift tests now run with lowering both enabled and disabled.
> 
> Exposes the feature via `zkc compile --lower-zkc-native` and updates IR string formatting to omit `+ 0` for `BIT_CONCAT` similarly to add/sub.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c610d53c7fa6c056d52e2e630b1781b0a0961ad9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/cmd/zkc/compile.go
```diff
@@ -59,6 +59,7 @@ func runCompileCmd[F field.Element[F]](cmd *cobra.Command, args []string, field
 	var (
 		// compiler config
 		config = codegen.DEFAULT_CONFIG.
+			LowerBitwise(GetFlag(cmd, "lower-bitwise")).
```
DavePearce: I'm not personally sure we need this flag, but it could be useful for debugging.

pkg/test/zkc_unit_test.go
```diff
@@ -359,71 +360,75 @@ func Test_ZkcUnit_Continue_01(t *testing.T) {
 // ===================================================================
 
 func Test_ZkcUnit_Bitwise_01(t *testing.T) {
-	checkZkcUnit(t, "zkc/unit/bitwise_01", field.BLS12_377, field.KOALABEAR_16)
+	checkZkcUnitWithLowerBitwise(t, "zkc/unit/bitwise_01", field.BLS12_377, field.KOALABEAR_16)
```
DavePearce: Ok, I see --- you're testing behaviour with and without the lowering.  Nice!

testdata/zkc/unit/bitwise_09.zkc
```diff
@@ -0,0 +1,40 @@
+pub input data(address:u8) -> (byte:u24)
+
+// chack that cast (a | b) == (cast a | cast b) ... and equivalent for and, xor, not
```
DavePearce: Minor typo

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+	"github.com/consensys/go-corset/pkg/zkc/vm/function"
+	"github.com/consensys/go-corset/pkg/zkc/vm/instruction"
+	"github.com/consensys/go-corset/pkg/zkc/vm/instruction/opcode"
+	"github.com/consensys/go-corset/pkg/zkc/vm/machine"
+	"github.com/consensys/go-corset/pkg/zkc/vm/word"
+)
+
+// LowerBitwise rewrites VM-level bitwise micro-instructions into CALLs to
+// helper functions. The helper modules are appended to the returned module
+// slice.
```
DavePearce, line 32: Would be useful to include the assumptions here.  Specifically, that this runs before vectorization and register splitting.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+		return bitwiseCall(id, t.Target, t.Sources, origWidth, p, registers)
+	default:
+		panic(fmt.Sprintf("unexpected non-bitwise opcode: %d", code.OpCode()))
+	}
+}
+
+// bitwiseCall emits a call to a bitwise helper module.  When origWidth equals p
+// (already a power of two) the call is direct with no temporaries.  Otherwise
+// each source is zero-extended from origWidth to p bits via Cast before the
+// call, and the p-wide result is truncated back to origWidth bits afterwards.
```
DavePearce, line 128: We'll have to think about how we optimise these casts in cases where they can actually be no-ops.  Anyway, for another day.

pkg/zkc/compiler/codegen/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+	insns = append(insns, instruction.NewCast(target, pResult, origWidth))
+
+	return insns
+}
+
+// bitwiseCallNot emits the instruction sequence for a NOT with non-power-of-2
+// width.  Unlike AND/OR/XOR, NOT flips the zero-padding bits when called on a
+// padded-up value, so the result must be masked back to origWidth before the
+// narrowing cast.
+func bitwiseCallNot[W word.Word[W]](
```
DavePearce: Actually, I think we can do better here.  Bitwise not can be implemented with subtraction directly like so:

```
input data(addr:u16) -> (x:u8)

fn main() {
   var v:u8 = data[0]
   printf "v=%b, ~v=%b\n", v, not(v)
}

fn not(a:u8) -> (r:u8) {
   r = 0xff - a
}
```

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+func nextPowerOfTwo(w uint) uint {
+	p := uint(1)
+	for p < w {
+		p <<= 1
+	}
+
+	return p
+}
+
+func lowerableWidth(registers []register.Register, target register.Id, bandWidth uint) (uint, bool) {
```
DavePearce, line 204: We don't need to consider the case of a native register here.  We are not allowed to call bitwise operations on native registers.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+		w = bandWidth
+	} else {
+		w = reg.Width()
+	}
+
+	if w == 0 {
+		panic(fmt.Sprintf("zero-width register: %s", reg.Name()))
+	}
+
+	return w, w&(w-1) == 0
```
DavePearce, line 218: Does that really work? `w&(w-1)` as a way of telling is power of two? 

.... so if not power of 2 then `w-1` always leaves high bit untouched ==> therefore, result not `0`.  

... Ok, nice.  I have not seen that trick before.

pkg/zkc/compiler/codegen/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+		width:    width,
+		arity:    arity,
+		constant: helperConstant(op, constant),
+	}
+
+	if id, ok := p.ids[key]; ok {
+		return id
+	}
+
+	// Sub-helpers (e.g. recursive NOT halves) are appended to p.items inside
```
DavePearce: didn't understand `recursive NOT halves` ?

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -0,0 +1,628 @@
...
+			agg = b.combineBit(key.opcode, agg, inp)
+		}
+
+		b.emit(instruction.NewIntAdd(out, []register.Id{agg}, zero))
+	} else {
+		// Recursive case.
+		half := key.width / 2
+
+		// Split the constant at generation time.
+		constBig := constant.BigInt()
```
DavePearce, line 356: There seems to be an issue with using constants which I did not figure out.  Specifically, this program:

```
fn and(a:u4, b:u4) -> (r:u4) {
   r = a & b & 0xe
}
```

Generates this decomposition:

```
fn and(u4 a, u4 b) -> (u4 r) {
[0]     r = $bit_and_u4(a, b) ; ret
}

...
```

You should maybe add a test for this case also.

DavePearce: Otherwise, the decomposition looks good.  And, I realised after looking at your code that you could support non-power-of-2 bitwise operations like this.  For example, `u3` subdivides into the `low` and `high` components e.g. `u1` and `u2`.

DavePearce: But, I guess that, provided we optimise the cast situation then upcasting to the next power of 2 will come at essentially no cost.

DavePearce (commented):
Ok, I think it is pretty much and looks good.  However, I found two issues (see comments for details): (1) problem with e.g. `a & b & 0xe`; (2) optimisation for bitwise not.  Also, maybe create a follow on issue for shift.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
letypequividelespoubelles: oh really, why can't we ? As it's feasible, I dont see why we would exlude it ...

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
letypequividelespoubelles: yes, and if power of two, equivalent to not, so anding it return 0

pkg/zkc/compiler/codegen/lower_bitwise.go
letypequividelespoubelles: oh wow never thought of this ! I'll do like this, it'll be way better !

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
letypequividelespoubelles: > Otherwise, the decomposition looks good. And, I realised after looking at your code that you could support non-power-of-2 bitwise operations like this. For example, u3 subdivides into the low and high components e.g. u1 and u2.

Hum ... I don't get your point. What I do is for a and b two u3, we generate:
`(a & b) as u4 = bit_and_u4( (a as u4) , (b as u4))`

letypequividelespoubelles: And I don't understand why the & 0xe has disappeared ?Where the optim came from ?

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
letypequividelespoubelles: yes ... could open a perf issue to remember it though

letypequividelespoubelles: https://github.com/Consensys/go-corset/issues/1737

pkg/zkc/compiler/codegen/lower_bitwise.go
letypequividelespoubelles: well, that's normal, I guess it's outdated comment :(

pkg/zkc/compiler/codegen/lower_bitwise.go
letypequividelespoubelles: implemented

pkg/zkc/compiler/codegen/lowerzkcnative/binarize_bitwise.go
```diff
@@ -0,0 +1,141 @@
...
+		op, target, sources, constant = t.OpCode(), t.Target, t.Sources, t.Constant
+	default:
+		return []instruction.Instruction{code}
+	}
+
+	if len(sources) <= 2 {
+		return []instruction.Instruction{code}
+	}
+
+	width := (*registers)[target.Unwrap()].Width()
```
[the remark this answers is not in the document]
DavePearce: Not technically a problem because native registers cannot perform bitwise operations.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_shift.go
```diff
@@ -0,0 +1,210 @@
...
+
+	a, n, out := b.inputs[0], b.inputs[1], b.output
+	width := key.width
+
+	zero := word.Uint64[W](0)
+	one := word.Uint64[W](1)
+	wmax := word.Uint64[W](uint64(width - 1))
+
+	zeroReg := b.newComputedNamed("zero", amtWidth)
+	wmaxReg := b.newComputedNamed("wmax", amtWidth)
```
DavePearce, line 127: Its actually annoying that we have to store this in a register.  We didn't have to do this in `zkasm` because the `skip_if` instruction allowed constants.  We will fix this (but not for this PR)

pkg/zkc/compiler/codegen/config.go
```diff
@@ -18,8 +18,9 @@ import "github.com/consensys/go-corset/pkg/util/field"
 // debugging, for example) should derive a custom Config via the chainable
 // setters below.
 var DEFAULT_CONFIG = Config{
-	field:     field.KOALABEAR_16,
-	vectorize: true,
+	field:          field.KOALABEAR_16,
+	lowerZkcNative: false,
```
DavePearce, line 22: The name is a bit too long for my liking.  Even just `lowerNative` or `lowerNatives` would be better IMHO

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -0,0 +1,538 @@
...
+
+	var zeroW W
+
+	mask := zeroW.SetBigInt(maskBig)
+	zero := word.Uint64[W](0)
+
+	maskReg := allocTmp(registers, width)
+
+	return []instruction.Instruction{
+		instruction.NewIntAdd(maskReg, nil, mask),
```
DavePearce, line 175: It would nice if we could write this in one instruction by having an alternative subtraction form which is `C - r1 - r2 -r3`, etc.

DavePearce (approved):
Overall, looks good to me (modulo the outstanding issue with constants to be fixed).

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
DavePearce: I'm not a fan of bitwise operations on native fields.  For example, what is `0x7F000000 | 0x01` under KOALABEAR?  Also, you'd roughly have this:

```

fn $bit_and_F(F arg1, F arg2) -> (?? out) {
        var lo1, hi1:u16
        var lo2, hi2:u16
        var rlo, rhi:u16
[0]     hi1::lo1 = arg1 ; hi2::lo2 = arg2 ; ... ; out = (F) rhi::rlo
```

Which works kind of, except that the cast to `out` can fail.  I think its better for force casting before doing the bitwise operation.

PR #1750, opened 2026-05-11
feat: lower SKIP_IF

### LINKED ISSUE #1736: feat: lowering comparisons in ZkC
## Summary

The ZkC language permits comparisons of the form `x < y` which have no direct correspondence at the AIR level.  Therefore, these need to be "compiled out" using subtraction.

## Example

In `zkasm` we typically implement comparisons directly using subtraction like do:

```
fn max256(arg1 u256, arg2 u256) -> (res u256) {
   var b u1
   var tmp u256
   b,tmp = arg1 - arg2
   if b!=0 goto exit1
   res = arg1
   return
exit1:
   res = arg2
   return
}
```

Here, we see that `arg1 < arg2` is implemented using `b,tmp = arg1 - arg2` where `b==1` implies `arg1 < arg2`.

letypequividelespoubelles:
> [!NOTE]
> **Medium Risk**
> Changes the ZKC codegen lowering pipeline and rewrites relational `SkipIf` semantics, which can affect correctness of compiled programs and introduces new width-handling logic in shift helpers.
> 
> **Overview**
> Adds a new lowering pass (`LowerComparisons`) that rewrites relational `SkipIf` conditions (`LT/GT/LTEQ/GTEQ`) into arithmetic-only instruction sequences (biased subtraction + sign-bit extraction), and wires this pass into compilation after bitwise lowering when `LowerZkcNative` is enabled.
> 
> Tightens temporary register allocation/naming in bitwise lowering helpers and fixes shift-helper comparison width mismatches by zero-extending shift amounts to a computed `wmax` width before emitting `LTEQ`.
> 
> Introduces new ZKC unit test vectors for VM `SkipIf` behavior (`skip_if_01` enabled; `skip_if_02`–`skip_if_05` added but currently commented out pending issue #1758).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12a09206b9689615919aca6ead33251a4215a4d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/zkc/compiler/codegen/lowerzkcnative/lower_comparison.go
```diff
@@ -0,0 +1,200 @@
...
+	si *instruction.SkipIf,
+	registers *[]register.Register,
+	bandWidth uint,
+) []vm.WordInstruction {
+	lhs, rhs, skipOnZero := normalizeRelational(si)
+
+	lhsWidth := resolveRegisterWidth(*registers, lhs, bandWidth)
+	rhsWidth := resolveRegisterWidth(*registers, rhs, bandWidth)
+
+	castBandWidth := lhsWidth
```
DavePearce: You can write `castBandWidth := max(lhsWidth, rhsWidth)` in Go.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_comparison.go
```diff
@@ -0,0 +1,200 @@
...
+			instruction.NewCast(aBase, lhs, castBandWidth-1),
+			instruction.NewBitConcat[W](biased, []register.Id{aBase, oneReg}),
+		}
+	}
+
+	subtractAnsDestruct := []vm.WordInstruction{
+		instruction.NewIntSub(diff, []register.Id{biased, bWide}, zero),
+		instruction.NewDestruct([]register.Id{lo, sign}, diff),
+		instruction.NewIntAdd(zeroReg, nil, zero),
+	}
```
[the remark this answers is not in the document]
DavePearce: These operations are supported in the field machine, even if they are not yet implemented.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_shift.go
```diff
@@ -132,7 +145,10 @@ func newShlHelper[W vm.Word[W]](key bitwiseHelperKey, selfID uint, amtWidth uint
 	b.emit(instruction.NewReturn())
 
 	// if n >= width: return 0
-	b.emit(instruction.NewSkipIf(opcode.LTEQ, n, wmaxReg, 2))
+	// Zero-extend n to wmaxWidth so both sides of LTEQ share the same register width.
+	nWide := b.newComputedNamed("nWide", wmaxWidth)
```
DavePearce: `nWide` is not a great name as it could clash with another column of the same name.  Remember these names are going to become column names, so this is not ideal.  You want instead a mechanism for creating "fresh" variable names, roughly of the form `$XYZ` where `XYZ` is some number that increments on every new name.  Likewise for the others above.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_comparison.go
```diff
@@ -0,0 +1,200 @@
...
+	if rhsWidth > castBandWidth {
+		castBandWidth = rhsWidth
+	}
+
+	castBandWidth++
+
+	zero := vm.Uint64[W](0)
+	one := vm.Uint64[W](1)
+
+	bWide := allocTmp(registers, castBandWidth)
```
DavePearce, line 119: So, at the moment, its pretty unpleasant in terms of the number of registers being created.  This is really coming from limitations in the `vm` instruction set.  For example, we should have an instruction "subtract with borrow" which allows us to do the subtraction directly rather than shifting it right by 1.  Also, better support for constants, no need for cast instructions, etc.

pkg/zkc/vm/lower.go
```diff
@@ -115,10 +115,7 @@ func (p wordToField[W, F]) lowerWordInstruction(wi WordInstruction) (fi FieldIns
 	case opcode.SKIP:
 		return wi.(*instruction.Skip)
 	case opcode.SKIP_IF:
```
DavePearce, line 117: Random why you removed the comment?  Its there just to clarify that we don't need to handle cases other than EQ/NEQ ... because they will cause a `panic`.  This actually wasn't original the case!!  Anyway, not a big deal.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_comparison.go
```diff
@@ -0,0 +1,200 @@
...
+	oneReg := allocTmp(registers, 1)
+	biased := allocTmp(registers, castBandWidth)
+	diff := allocTmp(registers, castBandWidth)
+	lo := allocTmp(registers, castBandWidth-1)
+	sign := allocTmp(registers, 1)
+	zeroReg := allocTmp(registers, 1)
+
```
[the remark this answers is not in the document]
DavePearce: This would be a problem only for registers with field type I think.  But such registers should not be permitted to be involved in comparisons.

DavePearce (approved):
Ok, some comments --- no deal breakers.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_comparison.go
letypequividelespoubelles: yep, do agree. But without a "substract borrow" vm instruction, I don't think we could do better. Raise an issue to create this new custom ?

DavePearce: Yes, agreed.

pkg/zkc/vm/lower.go
letypequividelespoubelles: hum ... I don't know why I modified this, I remember having rebase conflict ... I'll put it back

DavePearce (approved):
LGTM

PR #1783, opened 2026-05-15
feat(zkc): lower division

### LINKED ISSUE #1729: feat: lowering division / remainder in ZkC
_(see also #1723)_

## Summary

An important part of the constraint generation process is to lower certain "high-level" constructs into a more "low-level" representation. In this case, we are considering the division and remainder operations: `/` and `%`. These operations cannot be translated directly into constraints inline. Instead, they must be implemented as non-deterministic computations whose values are suitably constrained.

## Example

Consider the following `zkasm` program:

```
fn div(x u16, y=1 u16) -> (q u16, r u16) {
  var sum u32
  var wit u16
  var witsum u17
  ;;
  q, r, wit, sum, witsum = x / y
  return
}
```

This is compiled into the following uASM representation:

```
fn div(x u16, y u16) -> (q u16, r u16) {
      var sum u32
      var wit u16
      var witsum u17

[0]   q, r, wit = x / y ; 
           sum = r+(y*q) ; 
           witsum = 1+r+wit ; 
           skip sum != x 2 ;  // fail if sum != x
           skip witsum != y 1 ; // fail is witsum != y
           ret ; 
           fail
}
```

What we can see is that the instruction `q, r, wit = x / y` is turned into a non-determinstic computation with its return values being constrained appropriately after the fact.

letypequividelespoubelles:
> [!NOTE]
> **Medium Risk**
> Adds a new VM hint opcode and a lowering pass that rewrites `INT_DIV`/`INT_REM`, affecting core compiler/VM execution and field-lowering behavior; bugs here could change proof soundness or runtime semantics for division-heavy programs.
> 
> **Overview**
> Adds a new `LowerDivisions` lowering pass that expands `INT_DIV`/`INT_REM` into a `HINT_DIVISION` (`FieldHint`) plus arithmetic checks (including `r < y`) and wires it into the `lowerZkcNative` pipeline before comparison lowering.
> 
> Introduces the new `HINT_DIVISION` opcode and `FieldHint` instruction, implements execution in the word machine, and ensures word→field lowering preserves the hint instruction.
> 
> Improves codegen/vectorization robustness (stricter `endsInTerminator`) and refactors bitwise/comparison lowering to no longer depend on field bandwidth (panic if native regs appear).
> 
> Updates statement codegen to fold chains of constant divisors and adds new unit/invalid tests for division/remainder chains and related error cases; test harness now always runs a non-lowered variant and forces native lowering when constraint checks are enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e06a1a84343c2c06ecc1c92dc2b6063456ac612f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/zkc/compiler/codegen/vectorize.go
```diff
@@ -161,19 +161,37 @@ func prepareCode(code []VectorInstruction) []VectorInstruction {
 	return prepared
 }
 
-// endsInTerminator reports whether the last code unconditionally fixes the
-// next program counter (Jmp, Return or Fail).
+// endsInTerminator reports whether all micro-paths through codes terminate
+// without falling off the end: the last code must be a Jmp/Return/Fail, AND
+// no Skip/SkipIf anywhere in the vector has a skip target past the end (which
+// would create a second exit path not visible from the last instruction).
 func endsInTerminator(codes []Instruction) bool {
```
DavePearce, line 168: I'm a bit confused about the need for this change?  I guess it must mean that somewhere vectors are being generated which are not terminated because of skips.  Is this the right way to fix that?

pkg/zkc/vm/instruction/field/hint.go
```diff
@@ -0,0 +1,67 @@
...
+// registers are defined (written) by the prover without any polynomial
+// constraint; correctness is validated by subsequent arithmetic checks.
+type Hint struct {
+	Targets []register.Id
+	Sources []register.Id
+}
+
+// OpCode implementation for Instruction interface
+func (p *Hint) OpCode() opcode.OpCode {
+	return opcode.HINT_DIVISION
```
DavePearce, line 34: This seems OK for now.  Eventually we'll want more stuff here I guess for other hints.

pkg/zkc/vm/instruction/hint.go
```diff
@@ -0,0 +1,28 @@
...
+	finsn "github.com/consensys/go-corset/pkg/zkc/vm/instruction/field"
+)
+
+// FieldHint is a non-deterministic register assignment with no polynomial
+// constraint.  It marks the target registers as defined for the constancy
+// analysis without generating any equality.
+type FieldHint = finsn.Hint
+
+// NewFieldHint constructs a new hint instruction.
+func NewFieldHint(targets, sources []register.Id) *FieldHint {
```
DavePearce, line 26: I think this should go in `field.go` which is meant to collect all the "field instructions".

DavePearce: Ok, but then I see it is also used in the word machine.  So its probably a word instruction and goes into "word.go" with other word instructions.

DavePearce (approved):
Overall, I think it looks good.  I'm a bit confused about the change to `endsInTerminator()`, but that's it.  Removing need for explicit casts will eventually help simplify the generated `vm` instructions.

pkg/zkc/vm/instruction/field/hint.go
letypequividelespoubelles: yes, I think so.

pkg/zkc/vm/instruction/hint.go
letypequividelespoubelles: > So its probably a word instruction and goes into "word.go" with other word instructions.

hmm ... yes makes sense. I don't remember why I didn't want to have it in word.go ... I can do the change

pkg/zkc/compiler/codegen/vectorize.go
letypequividelespoubelles: >I guess it must mean that somewhere vectors are being generated which are not terminated because of skips. 

yes


> Is this the right way to fix that?

 I hope, but I wasn't able to find a better solution, 100% open to other solutions

Issue #1784, opened 2026-05-17
[bug]  static tables with 𝔽 values don't handle overflows


OlivierBBB:
The following test (with `<test-name>.koalabear_16.accepts` file that reduces to `{}`) fails when one uncomments the commented checks. Claude tells me that this is related to reduction mod the koalabear prime not taking place when e.g. comparing two `𝔽`'s
```go
// ===================================================================
// memory test: reading from static felt table
// ===================================================================

static felt_static(address:u8) -> (felt:𝔽) {
    0 as 𝔽,
    1 as 𝔽,
    0x7f000000 as 𝔽,   // represents -1
    0x7f000001 as 𝔽,   // should represent  0
    0x7f000002 as 𝔽,   // should represent +1
}

fn main() {
    // assignment from static table
    var felt_0:𝔽 = felt_static[0]
    var felt_1:𝔽 = felt_static[1]
    var felt_neg1:𝔽 = felt_static[2]
    var felt_zero:𝔽 = felt_static[3]
    var felt_pos1:𝔽 = felt_static[4]

    // algebra for fun
    if (felt_neg1 + felt_1) != felt_0 {
        fail
    }
    // if felt_zero != felt_0 {
    //     fail
    // }
    // if felt_pos1 != felt_1 {
    //     fail
    // }
    if (felt_neg1 + felt_1) != felt_0 {
        fail
    }
    if (felt_neg1 * felt_neg1) != felt_1 {
        fail
    }
}
```

## Minimal example

```go
// ===================================================================
// memory test: reading from static felt table, minimal example
// ===================================================================

static felt_static(address:u8) -> (felt:𝔽) {
    0 as 𝔽,
    0x7f000001 as 𝔽,   // should represent  0
}

fn main() {
    // assignment from static table
    var felt_0:𝔽 = felt_static[0]
    var felt_zero:𝔽 = felt_static[1]

    if felt_zero != felt_0 {
        fail
    }
}
```

DavePearce:
This example should not compile and, instead, should report a compile-time error here:

```
    0x7f000001 as 𝔽,
    ^^^^^^^^^
```

The issue lies in the handling of casts in `EvalConstant()`.  It needs to check that the casted value is in-bounds and, if not, return a `cast overflow` error or similar.

This is actually a bug in the implementation for #1778 rather than a separate issue.

PR #1785, opened 2026-05-17
[bug] writable field element memory + field element static tables

### LINKED ISSUE #1778: [bug] Field element memory can't be written to
The following defines a memory holding field element values.

```zig
memory felt_state(address:u8) -> (felt:𝔽)

fn write_to_state<felt_state>(address:u8, value:𝔽) -> () {
    felt_state[address] = value
}

fn read_from_state<felt_state>(address:u8) -> (value:𝔽) {
    value = felt_state[address]
    return
}
```

Reads work without issue, but writes don't.

```
panic: field element type has no fixed bitwidth: 𝔽 [recovered]
        panic: field element type has no fixed bitwidth: 𝔽
```

OlivierBBB:
This PR also fixes a widespread typo `flattern -> flatten`

---

> [!NOTE]
> **Medium Risk**
> Touches compiler codegen, typing, and constant evaluation for field elements; widespread `Flatten` rename is mechanical but touches many call sites.
> 
> **Overview**
> Fixes **field element (𝔽)** support in the ZKC compiler and VM path: writable **𝔽 memory** now allocates **native** registers (`math.MaxUint` bitwidth) in `codegen/statement.go`, and **static tables** / **casts** / assignments reject values that overflow the configured field modulus (typing + `ConstantEvaluator`, including `Cast`).
> 
> Renames the misspelling **`Flattern` → `Flatten`** (and `flattern*` helpers → `flatten*`) across corset source maps, module scopes, IR term simplification, ZKC codegen/lower, and related comments.
> 
> Other small changes: `IncrementalCompiler` defaults to **`field.KOALABEAR_16`**; memory-write panic text fixed to "unwritable"; new unit/invalid tests and **`KOALABEAR_CONFIG`** for 𝔽 overflow cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f39d59983463430b3440f441367983eba83375d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/zkc/compiler/codegen/evaluate.go
```diff
@@ -124,6 +129,10 @@ func EvalConstant(
 		return BitwiseXor(bitwidth, args...), ""
 	case *expr.Cast[symbol.Resolved]:
 		inner, _ := EvalConstant(e.Expr, definition, declarations, env)
+		if e.CastType != nil && e.CastType.AsField(env) != nil {
```
DavePearce: See comment left on #1784 --- I think there is a bug here in that it should be catching overflows.

DavePearce (changes requested):
This looks good to me, aside from the issue with cast implementation (see my note).

pkg/zkc/compiler/validate/typing.go
```diff
@@ -745,10 +745,25 @@ func (p *TypeChecker) typeCastExpression(_ Type, e *expr.Cast[symbol.Resolved],
...
+	}
+
 	// Sanity check whether the cast makes sense (or not).
-	if len(errors) == 0 {
-		errors = p.checkCastType(e.CastType, srcType, e.Expr)
+	errors = p.checkCastType(e.CastType, srcType, e.Expr)
+
+	// If casting to a field element, attempt to evaluate expression;
+	// overflowing the field modulus is disallowed and gets flagged;
+	if p.env.WellFormed(e.CastType) && e.CastType.AsField(p.env) != nil {
```
DavePearce: This looks odd to me.  It doesn't really make sense in my mind to do evaluate the constant at this point.  Especially, since the changes to `typeConst` should cover it.

testdata/zkc/mixed/felt_casting_01.zkc
```diff
@@ -0,0 +1,17 @@
+// ===================================================================
```
DavePearce, line 1: Tests involve native field elements should go in the `testdata/zkc/mixed` folder and be listed in the `zkc_mixed_tests.go` file.  These are for tests which involve native fields, since they require specialised testing.

pkg/test/util/check_valid.go
```diff
@@ -40,6 +40,14 @@ var (
 		words:       DEFAULT_WORDS,
 		constraints: false,
 		splitting:   false}
+	// KOALABEAR_CONFIG for tests requiring Koalabear to function (e.g. invalid/ tests
+	// triggering overflows in 𝔽 variable assignments
+	KOALABEAR_CONFIG = Config{
```
DavePearce: You don't need to add a new config here.  Firstly, you can already limit the `DEFAULT_CONFIG` using the `Fields()` option.  But, more importantly, the testing framework is already designed to handle the situation where a specific test only works for a specific field (or, conversely, where tests work but require different input/outputs for different fields).  Those tests in the `mixed` directly illustrate how to do this.  For example, you can have an `accepts` file with the name: `basic_01.koalabear_16.accepts`

testdata/zkc/unit/felt_casting_01.zkc
```diff
@@ -0,0 +1,17 @@
+// ===================================================================
+// felt casting test
+// ===================================================================
+
+fn main() {
+    var neg_one:𝔽 = 0x7f000000 as 𝔽 // ok
```
DavePearce: FYI you don't need to write `as 𝔽` here.  You can just write this instead:

```
var neg_one:𝔽 = 0x7f000000
```

DavePearce (dismissed):
Overall, this looks fine.  I made some minor tweaks to eliminate the constant evaluation happening in the type checker for casts.  Also, I dropped some casts from the tests (though not all).  And, added a few more tests.  I also moved the new tests into the `mixed` folder which is where tests involving native field elements should live.

pkg/zkc/compiler/codegen/evaluate.go
```diff
@@ -182,11 +182,18 @@ func (p ConstantEvaluator) evalFieldConstant(e Expr, definition bool) (res vm.Ui
 	modulus = modulus.SetBigInt(fmod)
 	//
 	switch e := e.(type) {
+	case *expr.Cast[symbol.Resolved]:
+		res, errorMessage = p.Eval(e.Expr, definition)
+		if res.BigInt().Cmp(fmod) >= 0 {
+			return res, fmt.Sprintf("cast overflows field modulus \"%s\"", p.field.Name)
+		}
+
+		return
```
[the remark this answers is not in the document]
DavePearce: addressed

DavePearce (approved):
LGTM

pkg/test/util/check_valid.go
OlivierBBB: Right, I forgot about that.

PR #1798, opened 2026-05-18
added test to reproduce ternary operator issue

lorenzogentile404:
> [!NOTE]
> **Low Risk**
> Low risk because changes are limited to new test coverage and fixtures; no production/compiler logic is modified.
> 
> **Overview**
> Adds a new benchmark-style test `Test_ZkcBench_Ternary_01` plus `ternary_01.zkc`/`.accepts` fixtures to reproduce a ternary-expression edge case where both branches are evaluated, triggering an overflow on a cast that would be avoided with an equivalent `if/else`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bedb1011e1c438103af9d7a718d818c6683975d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


letypequividelespoubelles:
in the end, in tge constraint, you'll have something like:
r = if cond then trueBranch else falseBranch
So you gave a register that is filled with trueBranchValue and a register that is filled with falseBranchValue and constraint
cond * (r - trueBranch) + ( 1 - cond) * ( r - falseBranch) (maybe not exactly this, but whatever, in the end you have constraint that mixes r, cond, trueBranc and falseBranch.

The aim is to have a trace + associated constraint that enforce to have the wanted execution. You can't change the constraint at run time. It's a compile time fixed thing.

So unless, I'm missing something, it's normal that this programm is failing at run time, and the compiler can't catch the issue at compile time, he can't know that you'll feed him with uncastable value.

So imho, there are no issues in corset

lorenzogentile404:
See comment in the test showing different behaviour of ternary operator and if, then, else.

DavePearce:
> So imho, there are no issues in corset

@letypequividelespoubelles No, there is definitely a problem here.  The given example above is similar to my infinite loop example, just a different variation on it.

DavePearce:
This was fixed and merged in #1810

PR #1799, opened 2026-05-18
feat: refacto flattenFixedArrays to fix and add whole array support

### LINKED ISSUE #1711: fix: execution failures for fixed arrays
## Summary

There are a number of outstanding execution failures arising from fixed arrays.  These failures occur under various scenarious, such as attempting to assign an entire array to a variable.

## Example 1

```
type ARR = [u8;4]

fn rev(xs:[u8;4]) -> (ys:ARR) {
    ys[0] = xs[3]
    ys[1] = xs[2]
    ys[2] = xs[1]
    ys[3] = xs[0]
}
```

This results in a `panic` because the function `lower.FlattenFixedArrays()` does not recognise `ys` as an array and, hence, does expand it into four variables.

## Example 2

```
fn f(before:[u8;4]) -> (r:[u8;4]) {
    r = rev(before)
}

fn rev(xs:[u8;4]) -> (ys:[u8;4]) {
    ys[0] = xs[3]
    ys[1] = xs[2]
    ys[2] = xs[1]
    ys[3] = xs[0]
}
```

This fails because of the assignment in `f()` to `r`.  Its not clear to me why this is failing.

## Approach

The problem with these examples stems from the `lower.flatten.go` implementation, which is not very robust.  It needs to be reworked in two says: (1) to make it clearer to understand; (2) to generalise it to support arbitrary composite types.

## NOTES

The unit tests `fixed_array_10` .. `fixed_array_12` are commented out because they exhibit these failures.

amkCha:
Closes #1711 
- fix case of bare assignment with source a ExternAccess (function call)
And add support for comparison with whole-array in 
- if conditions
- ternary operators (with true and false branch localAccess)

---

> [!NOTE]
> **High Risk**
> Large compiler lowering changes with control-flow PC remapping and in-place AST rewriting; incorrect branch targets or compare lowering would break generated programs silently.
> 
> **Overview**
> Refactors **fixed-size array lowering** out of `flatten.go` into `flatten_fixed_array.go` and extends what the compiler accepts before arrays are scalarized. `FlattenFixedArrays` no longer takes `srcmaps`, and lowering always runs per function (not only when an array variable is present).
> 
> **Assignments:** whole-array copy from another array still splits element-wise; **array results from function calls** (`ExternAccess`) are now expanded—single target `a = f()` and multi-target `a, b = f()` become per-element `lval.Array` targets on one assign.
> 
> **Comparisons:** typing adds **`==` / `!=` on matching fixed arrays** (`typeArrayComparison`). Lowering turns whole-array equality in **`if` conditions** into element-wise `IfGoto`s plus a final `Goto` for `==`, and **`remapPC`** fixes branch targets when expansion inserts extra instructions. **Ternaries** with whole-array `==` / `!=` become nested scalar ternaries, with **`cloneExpr`** so shared branch nodes are not remapped twice in place.
> 
> **Tests:** re-enables `fixed_array_10`–`12` (closes #1711), adds cases for multi-return assign, destructuring/bit-split, and array compare in `if`/ternary; **`fixed_array_16`** (destructuring) and **`fixed_array_22`** (compare via extern) stay disabled pending other issues.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a1b2a9baa496387c29a1ae60fe53665ee9443e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/zkc/compiler/lower/flatten.go
```diff
@@ -280,55 +280,95 @@ func countVarsOfKind(vars []variable.ResolvedDescriptor, kind variable.Kind) uin
...
+		for i := range tgt.size {
+			idx := *big.NewInt(int64(i))
+			result[i] = &stmt.Assign[symbol.Resolved]{
+				Targets: []lval.LVal[symbol.Resolved]{
+					lval.NewArray[symbol.Resolved](tgt.id, expr.NewConstant[symbol.Resolved](idx, 10)),
+				},
+				Source: &expr.ArrayAccess[symbol.Resolved]{
+					Id:       src.Variable,
+					Arg:      expr.NewConstant[symbol.Resolved](idx, 10),
+					Datatype: srcArr.DataType,
```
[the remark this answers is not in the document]
amkCha: at this stage, the source cannot be anything other than an array because the typing phase has been done and validated, else it errors

pkg/zkc/compiler/validate/typing.go
```diff
@@ -486,6 +486,10 @@ func (p *TypeChecker) typeComparison(e *expr.Cmp[symbol.Resolved], env VariableM
 	if wellFormed(anchor, p.env) && anchor.AsField(p.env) != nil && e.Operator.Fieldable() {
 		return p.typeFieldComparison(e, env, effects)
 	}
+	// whole-array equality (==, !=) on matching fixed arrays
+	if wellFormed(anchor, p.env) && anchor.AsFixedArray(p.env) != nil {
+		return p.typeArrayComparison(e, lhs, rhs, lerrs, rerrs)
```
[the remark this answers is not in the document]
DavePearce: I will have a look at this and see I can recreate it

pkg/zkc/compiler/validate/typing.go
```diff
@@ -503,6 +507,33 @@ func (p *TypeChecker) typeComparison(e *expr.Cmp[symbol.Resolved], env VariableM
...
 }
 
+// typeArrayComparison type-checks a comparison whose operands are fixed
+// arrays.  Only `==` and `!=` are supported; both sides must be fixed arrays
+// of equal size and element type.
+func (p *TypeChecker) typeArrayComparison(e *expr.Cmp[symbol.Resolved],
+	lhs, rhs Type, lerrs, rerrs []source.SyntaxError,
+) []source.SyntaxError {
+	// Only == and != make sense element-wise on whole arrays.
+	if !e.Operator.Fieldable() {
```
DavePearce, line 518: Curious name "fieldable".  I guess it means: something which can be used at the field level?

pkg/zkc/compiler/compiler.go
```diff
@@ -77,7 +77,7 @@ func Compile(field field.Config, files ...source.File) (ast.Program, source.Maps
 	errors = append(errors, validateProgram(program, field, srcmaps)...)
 	// Lower fixed-size arrays into flat local access registers
 	if len(errors) == 0 {
-		lower.FlattenFixedArrays(program, srcmaps)
+		lower.FlattenFixedArrays(program)
```
DavePearce: I was trying think about what the pros/cons are for doing the flattening right here.  One downside I can think of, is that we might want to see the form before this lowering is applied.  It might make more sense to put it at the beginning of `Compile()` in `codegen/compile.go`.  That way, when we run `zkc compile --ast` we see what was before, whilst `zkc compile --wir` shows us after arrays are expanded.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -32,25 +32,25 @@ import (
 
 // FlattenFixedArrays expands
 // fixed-size array variables into individual scalar variables.  A variable
-// arrayName of type uM[n] is replaced by n scalars arrayName$0 .. arrayName$(n-1),
+// arrayName of type [uM;n] is replaced by n scalars arrayName$0 .. arrayName$(n-1),
 // each of type uM.  Corresponding expr.ArrayAccess and lval.Array nodes are
 // rewritten to plain LocalAccess / lval.Variable references.
-func FlattenFixedArrays(program ast.Program, srcmaps source.Maps[any]) {
+func FlattenFixedArrays(program ast.Program) {
```
DavePearce: This file is getting big.  We should put `FlatterFixedArrays` in its own file, like `flattern_arrays.go`.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -280,77 +354,415 @@ func countVarsOfKind(vars []variable.ResolvedDescriptor, kind variable.Kind) uin
...
 
-		target := lval.NewArray[symbol.Resolved](lv.Ids[0], expr.NewConstant[symbol.Resolved](idx, 10))
+	return expr.NewCmp(op, lhs, rhs)
+}
 
-		result[i] = &stmt.Assign[symbol.Resolved]{
-			Targets: []lval.LVal[symbol.Resolved]{target},
-			Source:  access,
+// remapPC rewrites the PC of a statement from old to new PC space based on the recorded shifts.
+func remapPC(oldPC uint, pcMapping []PcMapping) uint {
```
DavePearce: So, yeah, I can't remember what we talked about this before.  But, there is actually another way we can do this now, which I guess its a better option because it avoids having to do the remapping:

```
fn f(arr:[u4;2]) -> (r:[u4;2]) {
   narr = arr
}
```

Could be translated like this:

```
fn f(arr:[u4;2]) -> (r:[u4;2]) {
   narr$0, narr$1 = arr$0, arr$1
}
```

DavePearce: I'm not sure its worth doing more on this in this PR though.  See my comments at at top.  Maybe for later.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -280,77 +354,415 @@ func countVarsOfKind(vars []variable.ResolvedDescriptor, kind variable.Kind) uin
...
+			inner = t
+			constBranch = cloneExpr(ifTrue)
+		}
+	}
+	//
+	return inner
+}
+
+// cloneExpr returns a deep copy of e.
+func cloneExpr(e expr.Resolved) expr.Resolved {
```
DavePearce: Using cloning like this is certainly something that I do, and that makes sense when required.  What you have here looks essentially fine.  But, there are some hidden issues.  Specifically, we are not retaining the source mapping information here.  **What does this mean?**  It means that syntax errors reported in a subsequent phase (`codegen.Compile()` in this case) may not work properly.  Minimal example to illustrate:

```
const TWO : u8 = 2

fn f(arr1:[u8;4],arr2:[u8;4], y:u8) -> (r:u8) {
   r = (arr1 == arr2) ? 0 : (y / 128 / TWO)
}
```

The good news is that it took me a while to find this.  I thought an example existed, but its not really a very concerning example (i.e. unlikely to arise in practice).  Ok, I found another (again not very concerning one):

```
fn f(arr1:[u8;4],arr2:[u8;4], y:u8) -> (r:u8) {
   r = (arr1 == arr2) ? 0 : (128*2)
}
```

The more concerning aspect for me, actually, is that I find generating syntax errors to be quite useful for debugging because they point to the source location where it comes from.

Anyway, in this case, it is (in theory) easy to resolve the problem with `cloneExpr()` --- we just copy over the source mapping information from the original node to the cloned copy of that node.  A method `source.Maps.Copy()` is provided for doing this, and you can see places where I've used it (e.g. see `lowerStatementExprs()` in later on in `flatten.go`).  I was going to try and do it, but I realise it would mean changing a lot of functions to pass the `srcmaps` down, and I didn't want to make such a big change on your PR.  **I don't know if I would bother at this stage** -- the cases are not that concerning (and again see my comments at the top of the file).

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -61,181 +61,255 @@ func FlattenFixedArrays(program ast.Program, srcmaps source.Maps[any]) {
...
 // list.  For scalar variables newBase is the single new ID.  For fixed arrays
-// newBase..newBase+size-1 are the individual element variables.
-type varMapping struct {
-	newBase uint
-	isArray bool
-	size    uint
+// newBase..newBase+size-1 are the individual element variables and elemType
+// is the canonical element type used by the lowering helpers.  elemType
+// is nil for non-array entries.
+type VarMapping struct {
```
DavePearce: This is definitely an improvement, and reflects more like what I was expecting.  We needed a mapping from the old variable identifies to the new variable identifiers ... done 👍

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -61,181 +61,255 @@ func FlattenFixedArrays(program ast.Program, srcmaps source.Maps[any]) {
...
-			//
+		base := uint(len(expandedVars))
+
+		if vType := v.DataType.AsFixedArray(env); vType != nil {
+			size := vType.Size.First()
+
 			hasArray = true
+
+			bitwidth, ok := data.BitWidthOf(vType, env)
+			if !ok {
```
DavePearce: This is very minor, and don't action this.  But, should someone ever wants to use an array of field elements `[]𝔽` then this will fail.  But we can address that if and when the time comes.

pkg/zkc/compiler/lower/flatten.go
```diff
@@ -61,181 +61,255 @@ func FlattenFixedArrays(program ast.Program, srcmaps source.Maps[any]) {
...
 // into individual indexed accesses (e.g. sum(items) becomes
 // sum(items[0], items[1], items[2])).  All expanded nodes use the original
 // variable IDs so that the subsequent rewriting phase can remap them.
 func expandFixedArrays(
-	fn *decl.ResolvedFunction, mapping []varMapping, env ast.Environment,
+	fn *decl.ResolvedFunction, varMapping []VarMapping, env ast.Environment,
 ) (expandedVars []variable.ResolvedDescriptor, expandedCode []stmt.Resolved, hasArray bool) {
+	expandedVars, hasArray = expandFnVariables(fn, varMapping, env)
+
+	if !hasArray {
```
DavePearce: Initially, I thought --- this makes sense.  If there is no array being used, then we avoid the cost of expanding.  But, then I thought ... hmmmm, there are situations when this might not be true.  Two examples:

```
input data(addr:u16) -> (r:[u8;4])

fn main() {
    if data[0] == data[1] {
       fail
    }
}
```
(this actually fails because the type is not expanded for the `input` memory --- I guess that's a use case for another day).

```
fn main() {
    if f(0) == f(1) {
       fail
    }
}

fn f(x:u8) -> (r:[u8;2]) {
   r[0] = x
   r[1] = x
}
```

This one actually generates code, but that code is invalid.  It should fail an internal validation check on the `vm` instructions (but I haven't implemented that check yet).

**Summary**: just because a function doesn't have a variable of array type, this does not mean it doesn't use an array!

pkg/zkc/compiler/validate/typing.go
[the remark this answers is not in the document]
DavePearce: Ok, cursor bot is correct.  Here is my test case:
```
fn main() {
   var x:[u8;2]
   var y:[u8;2]
   //
   x[0] = 0
   x[1] = 1
   y[0] = 0
   y[1] = 2
   // ASSERT: x != y
   f(x,y)
}

fn f(l:[u8;2],r:[u8;2]) {
    if id(l) == id(r) {
       fail
    }
}

fn id(x:[u8;2]) -> (r:[u8;2]) {
   r = x
}
```

Running with an empty input file (i.e. `{}`) and we see a machine panic when we should not.  Again, the generated code is invalid and should have triggered an internal validation check (but I didn't implement these yet).

DavePearce (approved):
So, as you will see I spent quite a while going through this carefully.  Overall, there is good news and bad news.  

- The **good news** is that you have significantly improved the `FlatternFixedArrays()` function and it now looks much more like what I was expecting.  Overall, things are much improved! **On the basis of that, I would suggest we merge what you have here and I have approved it for this reason**.

- Ok, so the **bad news** is that there are still a lot problems.  Thinking about it carefully, my feeling is that flattening at the AST level is the wrong place to do this, and that is the root cause.  It would be easier to flatten at the machine level.  This is tricky because it has to be done (I think) during `codegen.Compile()` stage.  I'll try to explain why I think this is the better place to do it, and give an idea of how I would approach it based on what we know now (and specifically thinking about `struct`s).

(**Why AST is the wrong place**) The big problem with lowering this particular feature at the AST level is that it requires, in some places, to introduce temporary variables and (as Francois already discovered) this is actually extremely difficult to do at the AST level.  The simplest possible example to illustrate would be this:

```
fn f(x:[u8;2]) {
   f(id(x))
}

fn id(x:[u8;2]) -> (r:[u8;2]) {
   r = x
}
```

Please ignore the fact this code is non-sensical --- its just illustrating a point.  The current translation of `f()` is the following:

```
fn f(u8 x$0, u8 x$1) -> () {
[0]     f(id(x$0,x$1))
[1]     return
}
```

We can see that `f(id(x$0,x$1))` doesn't make sense.  So how could we fix this?  By **introducing a temporary variable** and rewriting like so:

```
fn f(x:[u8;2]) {
   var tmp : [u8;2]
   tmp = id(x)
   f(tmp)
}
```

This now gives the correct translation which looks like this:

```
fn f(u8 x$0, u8 x$1) -> () {
        u8 tmp$0
        u8 tmp$1
[0]     tmp$0, tmp$1 = id(x$0,x$1)
[1]     f(tmp$0,tmp$1)
[2]     return
}
```

**BUT**: unfortunately the process of introducing temporaries at the AST level is simply too difficult/impossible to get right; instead, introducing them at the `vm` level works much better.  This is roughly the same issue we had with the two suggested approaches for the ternary operator, and why I stuck with doing it at the `vm` level.

(**Approach I would take**).  There is a general process at work here, namely that of taking a compound variable (i.e. a variable composed of multiple variables, such as a fixed array or struct) and flattening (or inlining?) it into one or more variables.  That is, we go from `var x:[u8;2]` to `var x$0:u8, x$1:u8` or, for `struct`s, it would be `var p:{x:u8,y:u8}` to `var p$x:u8, p$y:u8`.

Therefore, I would build into the `codegen.Compiler()` function a general-purpose mechanism for doing this, and which we can then easily configure for fixed arrays and structs.  Exactly how that mechanism would look is not entirely clear to me (though I could make suggestions) and we would need to work through the details.  Its just my intuition that this would be the right way to approach it.

pkg/zkc/compiler/validate/typing.go
amkCha: For this one, I reuse one of your methods 👌 so I didn't challenge the naming https://github.com/Consensys/go-corset/blob/main/pkg/zkc/compiler/ast/expr/cmp.go#L45 Lmk if you'd rather I change it

pkg/zkc/compiler/lower/flatten.go
amkCha: Ah, I did try to find a case, but couldn't so I kept the skip ! It's removed now

pkg/zkc/compiler/lower/flatten.go
amkCha: recorded here => https://github.com/Consensys/go-corset/issues/1820

pkg/zkc/compiler/lower/flatten.go
amkCha: Kept a note for the 1st example here => https://github.com/Consensys/go-corset/issues/1820

pkg/zkc/compiler/lower/flatten.go
amkCha: Ok got it, actually now that you've introduced a `Rewriter` I could also have an `Expander` for the 1st phase of the flattening process and pass down srcmap like this - I'll prob do that in the next PR as I also pass declarations and env in all expansion functions

amkCha: Recorded in the same issue as the others :ok_hand:

pkg/zkc/compiler/compiler.go
amkCha: Hm, that's actually not great to never have the unexpanded version .. hence the initial config I guess, where we wanted to have it expanded or not. Putting in in codegen would be better

PR #1806, opened 2026-05-19
fix: non duplicate bit helpers module were created

### LINKED ISSUE #1802: fix: constraint generation failure for bitwise operation
## Summary

Some kind of internal accounting issue leads to a constraint generation failure for bitwise and operation.

## Example

Minimal example to illustrate (taken from `bitwise_09.zkc`):

```
pub input data(address:u8) -> (byte:u24)

// check that cast (a | b) == (cast a | cast b) ... and equivalent for AND, XOR
// where width a < power of 2 < width of cast a

fn main() {
    var x:u24 = data[0]
    var y:u24 = data[1]

    var xcast:u35 = x as u35
    var ycast:u35 = y as u35

    var and:u35 = (x & y) as u35
    var andcast:u35 = xcast & ycast
    if and != andcast {
        fail
    }

    return
}
```

**Steps to reproduce**:

```
> zkc compile --field=BLS12_377 --lower-native --air bitwise_09.zkc
```

Gives the following error: `panic: module "$bit_and_u1" already declared`

**Steps to confirm**:

```
> zkc compile --field=BLS12_377 --lower-native --mir bitwise_09.zkc | grep "bit_and_u1"
(module $bit_and_u1)
...
(module $bit_and_u1)
...
```

letypequividelespoubelles:
> [!NOTE]
> **Medium Risk**
> Changes the bitwise lowering/codegen path to stop specializing helper modules on per-call constants, which could affect correctness/performance of generated circuits if any constant-handling assumptions are wrong. Scope is limited to bitwise helper generation and associated unit coverage.
> 
> **Overview**
> Fixes duplicate bitwise helper module creation by **removing constant-specific specialization** from `LowerBitwise`’s helper cache and helper generation.
> 
> `BIT_AND/OR/XOR` (and `BIT_SHL/SHR`) now call `helpers.ensure(op, width, arity)` without a constant, and `newDecomposedNaryHelper` is updated to seed the base case with the opcode identity (AND=1, OR/XOR=0) and reuse a single sub-helper for both halves; `zeroWord`/`helperConstant` and the constant field in the helper key are removed. A small test cleanup drops an outdated TODO in `basic_23`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b54f4ce399dc62e39a8416ca78c4f60107ef73b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
+++ b/pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
@@ -92,24 +92,29 @@ func lowerBitwiseCode[W vm.Word[W]](
 
 	switch t := code.(type) {
 	case *instruction.BitAnd[W]:
-		id := helpers.ensure(t.OpCode(), p, len(t.Sources), t.Constant)
+		// After BinarizeBitwise, any non-identity constant has been
+		// materialised as a source register, so we can drop the constant
+		// argument here: at the (possibly widened) helper width the original
+		// identity mask is redundant because the cast already zero-extends
+		// inputs.
+		id := helpers.ensure(t.OpCode(), p, len(t.Sources))
 		return bitwiseCall(id, t.Target, t.Sources, origWidth, p, registers)
 	case *instruction.BitOr[W]:
-		id := helpers.ensure(t.OpCode(), p, len(t.Sources), t.Constant)
+		id := helpers.ensure(t.OpCode(), p, len(t.Sources))
 		return bitwiseCall(id, t.Target, t.Sources, origWidth, p, registers)
 	case *instruction.BitXor[W]:
-		id := helpers.ensure(t.OpCode(), p, len(t.Sources), t.Constant)
+		id := helpers.ensure(t.OpCode(), p, len(t.Sources))
 		return bitwiseCall(id, t.Target, t.Sources, origWidth, p, registers)
 	case *instruction.BitNot[W]:
 		// Inline ~x as (MASK - x) directly in the caller; no helper module needed.
 		return bitwiseInlineNot[W](t.Target, t.Sources[0], origWidth, registers)
 	case *instruction.BitShl[W]:
-		id := helpers.ensure(t.OpCode(), origWidth, len(t.Sources), zeroWord[W]())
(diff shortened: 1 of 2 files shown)


pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
```diff
@@ -214,10 +214,9 @@ func lowerableWidth(registers []register.Register, target register.Id) (uint, bo
 }
 
 type bitwiseHelperKey struct {
-	opcode   instruction.OpCode
-	width    uint
-	arity    int
-	constant string
+	opcode instruction.OpCode
+	width  uint
+	arity  int
```
DavePearce, line 219: Ok, it confused me a lot whey removing the constant from the key solved this problem.  However, I see the issue was that `helperName` did not include the constant in the name --- hence you could have different keys generating modules of the same name.

DavePearce (approved):
One thing you actually could do is to enable the `Bitwise_09` test.  This currently fails on `KOALABEAR_16`, but we can restrict it to `BLS12_377` like this:

```
func Test_ZkcUnit_Bitwise_09(t *testing.T) {
	// TODO: duplicate module #1802
	checkZkcUnit(t, "zkc/unit/bitwise_09", util.DEFAULT_CONFIG.Constraints(true).Fields(field.BLS12_377))
}
```

And then change the `TODO` to higlight register splitting as the outstanding problem with that test.

pkg/zkc/compiler/codegen/lowerzkcnative/lower_bitwise.go
letypequividelespoubelles: yep. The constant was coming from deduplicating an u25 & u 25 from a u26 & u26 cascade of modules (first implementation, I wasn't inlining the casting). Now I'm inlining the casting, it creates duplicates modules (that are in fact the same), and for the u1 modules, as you don't have the constant stuff two and_u2 where trying to create the same module. So rm the module helper duplicate fix the issue (and brings a huge perf as well :) )

Issue #1800, opened 2026-05-18
feat: constraint generation for Read Only Memory (ROM)


DavePearce:
## Summary

Currently, no constraints are generated for `input` memories (i.e. ROMs) in ZkC.  Therefore, the purpose of this feature is to add initial support for generating such constraints.

## Example

Consider the following very simple ZkC program:

```
input rom(address:u16) -> (byte:u8)
```

We can view the MIR constraints generated for this file using the following command:

```
> zkc compile --mir test.zkc
```

And this currently gives the following:

```
(module rom)

(inputs (address u16 0x0))
(outputs (byte u8 0x0))
```

The point being: there are no constraints generated.

## Assumptions

A read-only memory maps consecutive addresses to values, starting from address `0`.  For our example above, an initial set of constraints might be something like this:

```
;; address starts from 0
(defconstraint rom$init  (:domain {0}) (== 0 address))

;; address increments exactly by one
(defconstraint rom$inc (address == (+ 1 (prev address)))
```

The question is: what additional constraints do we need?  In particular, we at least need a mechanism for handling padding.

## Approach

For the purposes of this PR, we'll assume exactly one address line and zero or more data lines.  This simplifies the logic for incrementing `address` in our constraints (this can be done in a subsequent PR).  In the file `pkg/zkc/constraints/translator.go` we have the following:

```
func translateReadOnlyMemory[F field.Element[F]](_ schema.ModuleId, fm vm.InputOutputMemory[F]) mir.Module[F] {
	var (
		mod  *schema.Table[F, mir.Constraint[F]]
		name = trace.ModuleName{Name: fm.Name(), Multiplier: 1}
	)
	// Initialise module
	mod = mod.Init(name, false, true, false, fm.IsNative(), false, 0)
	// Add all registers
	mod.AddRegisters(fm.Registers()...)
	// TODO: implement ROM constraints
	return mod
}
```

In particular, the `TODO` marks the place where constraints need to be generated.  Looking at how constraints are generated for functions, particularly in `initMultiLineFraming()`, should be helpful in understanding how to actually construct the constraint objects.

DavePearce:
@OlivierBBB FYI looks like you are completing this issue as part of #1834.

Issue #1567, opened 2026-03-09
feat: mutation testing for ZkC


DavePearce:
_(**NOTE**: This is really a placeholder to accumulate knowledge around what exactly needs to be checked)_

## Summary

In order to test soundness of the ZkC compiler, we need to employ a mutation testing tool.  The goal of this tool is to generate traces which are **almost correct**, but not quite.  That is, traces which should be rejected by the generated constraints (and, hence, if they are not then we have a problem to fix).  

(**Almost Correctness**) The challenge with this tool is that we want to test the presence of individual constraints.  Thus, we need to generate traces that only incorrect in a very very minimal fashion.  To better understand this, consider attempting to test a program like `x = y + 1 ; z = x` --- specifically, to check the value for `x` is deterministic.  Suppose we take a valid trace for our program: 

```
x | 2 | 
y | 1 |
z | 2 |
```

And now simply mutate the the value of `x` randomly (e.g. say so `x==3`).  The problem with this is that constraints generated for the statement `z = x` will also fail.  Thus, we are not necessarily testing the constraints generated for our target statement `x = y + 1`.  Instead, we need to mutate the value of `x` at the point of the assignment, and then continue execution from that point so ensure all subsequent constraints are satisfied.  An example trace would be:

```
x | 3 |
y | 1 |
z | 3 |
```

Here, we can see the constraint generated for `z = x` will be satisifed, but that for `x = y + 1` will not.

## Notes

Some other aspects of mutation testing:

* **Casts / Arithmetic Overflow**. Consider a program `var x:u16, y:u8 ; x = y + 1` where `y==255`.  This program should be rejected due to the arithmetic overflow.  However, in fact, it will `panic` during trace execution preventing us from generating an invalid test.  Instead, we need to run a version of the executor which allows overflow to occur.  
* **Memory Reads / Writes**.  A similar issue to above exists for memory reads/writes, such as to write-once memory.  We need to be able to bypass the `panic` that would arise during execution in order to generate a faulty trace.
* **Trace Splicing**.  It is also possible to generate incorrect traces by splicing together valid traces (e.g. by looking for splice points which could be rows with identical or nearly identical values).

Issue #1543, opened 2026-03-02
feat: support `struct` types in ZkC


DavePearce:
_(this is a follow-on issue from #1536 and #1537)_

## Summary

The goal of this feature is to add `struct` types to the ZkC language.  For example, we could write something like this:

```
type Point = struct { u16 x, u16 y }

function getX(Point p) -> (u16 x) {
   x = p.x
   return
}
```

Assuming that #1536 and #1537 have been completed, this should be relatively straightforward to implement.  The approach is very roughly similar to that for fixed-size arrays.  An interesting question is whether or not to support some kind of _destructuring_ syntax --- but, eitherway, this could be added later in a subsequent issue.

amkCha:
Handle this case, reminder 

```
fn fg(i:u32) -> (r:[u32;2]) {
   r[0] = i
   return
}
````

PR #1834, opened 2026-06-02
[feat] Enforce semantics for write once memory (wom)

### LINKED ISSUE #1560: feat: enforce semantics for Write Once Memory (WOM)
## Summary

The semantics of a "write once memory" are exactly what it says on the tin.  You cannot write more than once to the same address in a WOM.  Attempting to do this should result in a machine panic.  But, currently, the ZkC interpreter does not prevent this and allows invalid executions to continue.

## Example

The following ZkC program is invalid and should result in a machine panic:

```
output out(address:u16) -> (byte:u8)

fn main(){
   out[0] = 1
   out[0] = 2 // <-- INVALID
}
```

With the following `json` input file:

```
{}
```

Then, currently, running this program gives the following:

```
> zkc exec test.json test.zkc
out = 0x02
```

Clearly, this violates the intended semantics of a "write once memory".

## Approach

Let's consider two example programs:

**Program A**:
```
output out(address:u16) -> (byte:u8)

fn main(){
   out[0] = 1
   out[1] = 2 
}
```

**Program B**:
```
output out(address:u16) -> (byte:u8)

fn main(){
   out[1] = 2
   out[0] = 1 
```

The implementation for write-only memory lives in `pkg/zkc/vm/internal/memory/wom.go`.  There are two different ways this implementation could be updated to enforce a write-once semantics:

1. (**strict**) This approach requires every write to an output memory is for the next consecutive address.  This would mean that "Program A" above is valid, but "Program B" is not.
2. (**non-strict**) This approach is more relaxed and simply requires that writes to the same address with different values result are not permitted.

The difference between these two approaches comes down to the complexity of enforcing their semantics.  Basically, option (1) is easier to implement than option (2).

### Implementation for Option (1)

In this option, its pretty straightforward.  Within our implementation of `memory.WriteOnce` we simply check that the write address matches the current size of the `data` array and, if not, cause a machine panic.

### Implementation for Option (2)

This is harder to implement, but of course gives rise to more flexible implementations for constructing output.  Its not clear whether this flexibility is needed or not.  The rough idea would be this: alongside the existing `data` array we additionally maintain an `written []bool` array of the same size.  This identifies addresses which have already been written.  Thus, when a write occurs at `address`, we check whether `written[address]==true` or not.  If it does, we issue a machine panic.

OlivierBBB:


pkg/test/zkc_unit_test.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/consensys/go-corset/pkg/zkc/vm"
 )
 
+func Test_ZkcUnit_Example(t *testing.T) {
+	checkZkcUnit(t, "zkc/unit/_example", util.DEFAULT_CONFIG)
```
DavePearce: Why does the test begin with an underscore?  That's not consistent with others.  Something like `basic_XX` or `ram_XX` would be consistent.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -83,7 +83,8 @@ func (p *StmtCompiler) compileStatement(pc uint, mapping []uint, s Stmt) VectorI
 //
 // > struct tmp { x u32, y u32 }
 // > ...
-// > var t tmp > tmp = f(...)
+// > var t tmp
+// > tmp = f(...)
```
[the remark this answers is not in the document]
DavePearce: Agreed, this is a typo

pkg/zkc/compiler/parser/environment.go
```diff
@@ -50,76 +50,76 @@ func EmptyEnvironment() Environment {
 // Clone constructs a clone of this environment, such that variables declared in
 // the clone will not clash with those declared elsewhere.  The inLoop parameter
 // indicates whether the cloned environment is inside a loop.
-func (p *Environment) Clone(inLoop bool) Environment {
+func (env *Environment) Clone(inLoop bool) Environment {
```
DavePearce, line 79: Not really sure what the motivation for this change is?  Did claude suggest this or something?  `p` is for "pointer", and is generally what is used in the code base.  Anyway, not an issue ... just not sure where its coming from.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
+	var name = trace.ModuleName{Name: fm.Name(), Multiplier: 1}
+	return translateMemoryCommon(ctx, fm, name)
+}
+
+// Write once memory and read only memory are equivalent on the constraints level
+func translateWriteOnceMemory[F field.Element[F]](
+	ctx schema.ModuleId, fm vm.InputOutputMemory[F]) mir.Module[F] {
+	var name = trace.ModuleName{Name: fm.Name(), Multiplier: 1}
+	return translateMemoryCommon(ctx, fm, name)
+}
```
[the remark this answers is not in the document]
DavePearce: I think this actually reflects what we expect to happen: ROMs and WOMs are the same.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
-	// Add all registers
-	mod.AddRegisters(fm.Registers()...)
-	// TODO: implement WOM constraints
-	return mod
-}
 
-func translateReadWriteMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory[F]) mir.Module[F] {
+	memoryModule.AddRegisters(register.NewComputed("timestamp_read", timestampWidth, padding))
+	memoryModule.AddRegisters(register.NewComputed("timestamp_write", timestampWidth, padding))
+	memoryModule.AddRegisters(register.NewComputed("timestamp_delta", timestampWidth, padding))
```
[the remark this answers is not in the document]
DavePearce: What copilot is saying here is that we (at some point) will need to think about trace generation as well.  But that could be a separate PR.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
-	// Initialise module
-	mod = mod.Init(name, false, true, false, fm.IsNative(), false, 0)
-	// Add all registers
-	mod.AddRegisters(fm.Registers()...)
-	// TODO: implement WOM constraints
-	return mod
-}
 
-func translateReadWriteMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory[F]) mir.Module[F] {
+	memoryModule.AddRegisters(register.NewComputed("timestamp_read", timestampWidth, padding))
```
DavePearce: Could we consider shorter names here?  `stamp_read` is at least better and consistent with the original arithmetisation.  I would go further and have just `r_stamp`.  Not really a big issue tbh, and can be addressed later (if it causes a problem).  Its just that we'll see these names used a lot when debugging the constraints, etc.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
+}
+
+func translateMemoryCommon[F field.Element[F]](
+	ctx schema.ModuleId, fm vm.Memory[F], name trace.ModuleName) mir.Module[F] {
 	var (
-		mod  *schema.Table[F, mir.Constraint[F]]
-		name = trace.ModuleName{Name: fm.Name(), Multiplier: 1}
+		memoryModule   *schema.Table[F, mir.Constraint[F]]
+		padding        big.Int
+		timestampWidth = uint(32)
```
DavePearce: I want to be clear here: **it is a bug to used a fixed bitwidth for the stamps**.  We should talk about how to resolve this (could be a separate PR even).  I have previously suggested syntax for expressing the timestamp width in `ZkC`.  Something like this:

```
memory ram(u32; address: u16) -> (byte:u8)
```

Here, the `u32;` at the start of `ram` declaration would be configuring the timestamp width.  What this actually is saying is that this memory can have a trace with at most `u32` elements.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
 	return mod
 }
 
-func translateReadOnlyMemory[F field.Element[F]](_ schema.ModuleId, fm vm.InputOutputMemory[F]) mir.Module[F] {
+func translateReadOnlyMemory[F field.Element[F]](
+	ctx schema.ModuleId, fm vm.InputOutputMemory[F]) mir.Module[F] {
+	var name = trace.ModuleName{Name: fm.Name(), Multiplier: 1}
+	return translateMemoryCommon(ctx, fm, name)
```
DavePearce: **NOTE.**  What I was hoping for in issue #1800 is that you add constraints to address ROM and WOM using lookups, along with a very simple set of constraints to ensure the address lines are incremented by one at each stage.  But, instead, you have jumped straight into the hard case of trying to distribute them across shards.   Generally, I always recommend doing the easy approach first and then progressing onto the harder issue.  The key benefit of the simple approach is that you could run it and test it now.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
+		if uint(i) < fm.Geometry().AddressLines() {
+			addressWidth += l.Width()
+		} else if uint(i) < fm.Geometry().AddressLines()+fm.Geometry().DataLines() {
+			valueWidth += l.Width()
+		}
+	}
+
+	// var address = register.NewId(memoryModule.Width() + 0)
+	// var valueRead   = register.NewId(memoryModule.Width() + 1)
+	memoryModule.AddRegisters(register.NewComputed("address", addressWidth, padding))
```
DavePearce: It doesn't make sense to add registers for the `address` and `data` (i.e. `valueRead`) lines as these are already part of the modules definition (i.e. they have already been added).

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
+	)
+
+	// ================================================
+	// constraints
+	// ================================================
+
+	// (non padding) rows are either created during standard execution (exec ≡ true)
+	// or during the finalization phase (finl ≡ true)
+	flagExclusivity := mir.NewVanishingConstraint("flag_exclusivity", ctx, util.None[int](),
+		mirc.Product([]Expr[F]{exec, finl}).Equals(zero).AsLogical())
```
DavePearce: I spent time looking at the use of multiplication here.  I prefer to use logical AND, as this is the point of MIR.  My version was this:

```
flagExclusivity := mir.NewVanishingConstraint("flag_exclusivity", ctx, util.None[int](),
		exec.Equals(zero).And(finl.Equals(zero)).AsLogical())
```

However, this does generate two constraints instead of 1 (i.e. as it doesn't implement it using `*`).  So, sticking with what you have here makes sense 👍

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
+	//
+	// exec may not be nondecreasing; the (exec, finl) pair may look like so :
+	//
+	//  ¹ ┼       ┌─────┐∙∙∙∙∙∙   ( ∙ ≡ finl)
+	//    │       │     │
+	//  ⁰ ┴  ─────┘∙∙∙∙∙└──────   ( ─ ≡ exec)
+	flagMonotony1 := mir.NewVanishingConstraint("finl_monotony", ctx, util.None[int](),
+		mirc.If(finlPrev.NotEquals(zero), finl.Equals(one)).AsLogical())
+	flagMonotony2 := mir.NewVanishingConstraint("exec+finl_monotony", ctx, util.None[int](),
+		mirc.If(mirc.Sum([]Expr[F]{execPrev, finlPrev}).NotEquals(zero),
```
DavePearce: Again here we could use logical OR instead of `Sum`.  My version was this:

```
flagMonotony3 := mir.NewVanishingConstraint("exec+finl_monotony2", ctx, util.None[int](),
		mirc.If(mirc.Or(execPrev.NotEquals(zero), finlPrev.NotEquals(zero)),
			mirc.Sum([]Expr[F]{exec, finl}).Equals(one)).AsLogical())

```

But, yeah, again it does generate two constraints instead of one.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
+			mirc.Sum([]Expr[F]{exec, finl}).Equals(one)).AsLogical())
+
+	// we want to prove WT - RT = 1 + ΔT (which forces WT > RT given that ΔT is ≥ 0)
+	// instead we prove WT = RT + 1 + ΔT
+	timestampMonotony := mir.NewVanishingConstraint("timestamp_monotony", ctx, util.None[int](),
+		mirc.If(exec.NotEquals(zero), wTime.Equals(rTime.Add(dTime, one))).AsLogical())
+
+	// var isImmutable bool
```
DavePearce: I'm not sure if this is code you want to keep for later so that it can be enabled again?  If so, seems fine since we cannot fully implement this solution without send/receive.  But, if its junk, then remove please.

testdata/zkc/unit/_example_0.zkc
```diff
@@ -0,0 +1,24 @@
+memory RAM(address :u4) -> (value :u4)
```
DavePearce: This test is not used anywhere ... ?

ram.md
```diff
@@ -0,0 +1,134 @@
+## RAM module
```
DavePearce: I guess this is like some working notes.  It doesn't belong in the `testdata/zkc/unit` directory.  Maybe put it in `docs/` where eventually we merge it into `docs/ZKC_ARCHITECTURE_CONSTRAINTS.md`.

rom.md
```diff
@@ -0,0 +1,94 @@
+## ROM module
```
DavePearce: ditto re location and also for the `wom.md` file.

pkg/zkc/constraints/translator.go
```diff
@@ -92,43 +92,144 @@ func translateStaticMemory[F field.Element[F]](_ schema.ModuleId, m vm.InputOutp
...
-	mod = mod.Init(name, false, true, false, fm.IsNative(), false, 0)
-	// Add all registers
-	mod.AddRegisters(fm.Registers()...)
-	// TODO: implement WOM constraints
-	return mod
-}
 
-func translateReadWriteMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory[F]) mir.Module[F] {
+	memoryModule.AddRegisters(register.NewComputed("timestamp_read", timestampWidth, padding))
+	memoryModule.AddRegisters(register.NewComputed("timestamp_write", timestampWidth, padding))
```
DavePearce: Its also strange to me that you would have a timestamp written for a ROM --- this doesn't make sense.

DavePearce: in fact, it doesn't make sense to have timestamps for `ROM` or `WOM` at all.  Yes for `RAM`, but no for the others.

pkg/zkc/constraints/translator.go
DavePearce: You can see this by trying to compile this file:

```
input in(address:u16)->(data:u8)
```

with e.g. `zkc compile --air test.zkc`

DavePearce: Likewise, if you compile it with `--mir` and look at the output, you will see this:

```
(inputs
   (address u16 0x0)
)

(outputs
   (data u8 0x0)
)

(computed
   (timestamp_read u32 0x0)
   (timestamp_write u32 0x0)
   (timestamp_delta u32 0x0)
   (address u16 0x0)                ;; <----- duplicate
   ...
)

...
```

DavePearce (commented):
Overall, it is good to see that you are getting the hang of how to write MIR constraints within the ZkC compiler.  There is some minor misunderstandings around the columns `address` and `valueRead` (i.e. since the necessary columns are already added from the memory declaration in ZkC).

However, some things about this PR also confuse me:

* (**Which Issue?**) One of the strange things about this PR is that it does not address issue #1560.  It is more like a PR for #1800 but anticipating the use send/receive instead of just lookups (which is what I was expecting).
* (**Timestamps**) It is confusing me why you are adding read/write timestamps for ROM/WOMs.  We don't need timestamps for these memory types (and, in fact, we cannot have them under the current design). Timestamps are intended specifically for accessing RAM.

To me, it feels more like you've been writing the constraints for RAM here instead those for ROM/WOM.  For example, `timestamp_monotony` is what I expect to see for RAM not for ROM/WOM.

pkg/zkc/constraints/translator.go
OlivierBBB: This whole block was removed for now.

docs/ZKC_STRUCTS.md
```diff
@@ -0,0 +1,386 @@
+# Structs in ZkC — Design
```
DavePearce: This file is not part of this issue, but seems related to #1543.

pkg/ir/mir/constraint.go
```diff
@@ -64,13 +64,25 @@ func NewLookupConstraint[F field.Element[F]](handle string, targets []LookupVect
 	return Constraint[F]{lookup.NewConstraint(handle, targets, sources)}
 }
 
+// NewSendConstraint creates a new "send to logUpBus" constraint; it doesn't take into account
```
DavePearce: Also, not really part of the issue --- but I guess its fine for now.

pkg/test/util/check_valid.go
```diff
@@ -443,13 +443,17 @@ func checkExpectedOutputs[W vm.Word[W]](outputs map[string][]W, wm vm.Core[W]) [
...
 		if output, ok := outputs[m.Name()]; ok {
-			if c := array.Compare(output, m.Contents()); c != 0 {
-				var (
-					expected = hex.EncodeToString(vm.EncodeBytes(output, m.Geometry()))
-					actual   = hex.EncodeToString(vm.EncodeBytes(m.Contents(), m.Geometry()))
-				)
-
-				errors = append(errors, fmt.Errorf("incorrect output (expected 0x%s, actual 0x%s)", expected, actual))
+			// Compare canonical byte encodings rather than the raw cell arrays:
+			// a memory whose live cell-count is odd encodes the same bytes as an
```
DavePearce, line 447: Ok, after thinking about, this seems pretty reasonable!

pkg/test/zkc_unit_test.go
```diff
@@ -250,6 +250,39 @@ func Test_ZkcUnit_Memory_01(t *testing.T) {
...
 }
 
+func Test_ZkcUnit_AccessOnceMemory_01(t *testing.T) {
+	// Multi-address-lane access-once memory: a read-only ROM and a write-once
+	// WOM, exercising the access bit and at_flag carry columns end-to-end.
+	checkZkcUnit(t, "zkc/unit/access_once_memory_01", DEFAULT_UNIT_CONFIG)
+}
+
+// DEFERRED: access_once_memory_02 is a multi-lane double-write that must be
+// REJECTED. The interpreter rejects it (WriteOnce.Write), but the harness can't
```
DavePearce, line 260: I guess the question is why we skip rejects tests -- I cannot remember the reason.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -47,6 +47,8 @@ type StmtCompiler struct {
 	errors      []source.SyntaxError
 	// quiet suppresses printf output
 	quiet bool
+	// fastMode disables constraint-only rewrites not required by the vm.
+	fastMode bool
```
DavePearce, line 51: This field is not required, as its not used.  I guess it might be a left over artifact of a rebase?

pkg/zkc/constraints/translator.go
```diff
@@ -112,34 +126,239 @@ func translateReadOnlyMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory
...
-	return mod
+
+	// ================================================
+	// ACCESS bit constraints
+	// ================================================
+
+	// We will impose the following:
+	//
+	//	- if ACCESS[i] = 0
+	//		- Then []ADDRESS[i] ≡ 0
```
DavePearce: What is this ... temporal logical notation?  I guess its trying to say: if `ACCESS[i]=0` then `ADDRESS[j]=0` for all `i <= j` --- is that right? Better to write (or at least summarise) in English.

pkg/zkc/constraints/translator.go
```diff
@@ -112,34 +126,239 @@ func translateReadOnlyMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory
...
+		//		- [a]ADDRESS[i]   =     [a]ADDRESS[i-1] for 0 ≤ a < k
+		//		- [k]ADDRESS[i-1] ≠ max_value_k
+		//		- [k]ADDRESS[i]   = 1 + [k]ADDRESS[i-1]
+		//		- [b]ADDRESS[i-1] = max_value_b for k < b < L
+		//		- [b]ADDRESS[i]   = 0           for k < b < L
+		//
+		// where L ≡ len([]ADDRESS) and max_value_x is the maximum integer of
+		// bitwidth that of the register [x]ADDRESS.
+		//
+		// **Note.** [0]ADDRESS is the *most* significant limb
```
DavePearce: That's right.  In my branch for register splitting, this is how it works now.

pkg/zkc/constraints/translator.go
```diff
@@ -112,34 +126,239 @@ func translateReadOnlyMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory
...
+		accessBitVanishesInPadding,
+		accessBitMonotony,
+	)
+
+	// ================================================
+	// []ADDRESS constraints
+	// ================================================
+
+	switch {
+	case isMultiLaneAddress:
```
DavePearce: This case is way too long.  You need to break it up into one or more sub-functions.

pkg/zkc/constraints/translator.go
```diff
@@ -112,34 +126,239 @@ func translateReadOnlyMemory[F field.Element[F]](_ schema.ModuleId, fm vm.Memory
...
+		for k := range addrRegs {
+			prevAddrRegs[k] = mirc.Variable[register.Id, Expr[F]](register.NewId(uint(k)), addrRegs[k].Width(), -1)
+			currAddrRegs[k] = mirc.Variable[register.Id, Expr[F]](register.NewId(uint(k)), addrRegs[k].Width(), 0)
+		}
+
+		// ================================================
+		// at_flag constraints
+		// ================================================
+
+		// we add binary '@k' flags to help locate the address register that
```
DavePearce: This is a random suggestion, but you might find it interesting to look at the existing implementation for lexicographic sorting in `pkg/ir/air/gadgets/lexicographic_sort.go`.  There is a lot of similarity here.

pkg/zkc/vm/internal/memory/geometry.go
```diff
@@ -56,6 +56,11 @@ func (p Geometry[W]) AddressLines() uint {
 	return p.numInputs
 }
 
+// IsMultiLineAddress reports whether the address spans more than one limb.
+func (p Geometry[W]) IsMultiLineAddress() bool {
```
DavePearce, line 60: The name of this function is confusing.  Its not "multi line" but "multi lane" I imagine.

pkg/zkc/vm/internal/memory/wom.go
```diff
@@ -22,6 +24,29 @@ import (
 // exactly what they are typically used for).
 type WriteOnce[W util.Uinter64] struct {
 	StaticArray[W]
+	writtenToAddresses map[uint64]bool
```
DavePearce: Please don't use a `map` --- its not efficient for a WOM.

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -53,10 +53,13 @@ func (p *FullObserver[W, I, M]) Initialise(machine M) {
 			// Static reference tables (e.g. the $range_un range-check tables) are
 			// not populated by execution, so their full contents must be seeded
 			// here; otherwise lookups into them would see an empty target.
-			p.trace[i] = initialiseROM(&m.ReadOnly)
+			p.trace[i] = initialiseAccessOnceMemory(&m.ReadOnly)
```
DavePearce: This does not seem right.  Static reference tables should not generate modules --- they are fixed as part of the constraints themselves.  How a lookup works with a static reference table is something to be discussed with the prover team.

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -127,18 +130,23 @@ func (p *FullObserver[W, I, M]) traceModule(m machine.Module, states []State[W],
 	builder array.Builder[util_word.BigEndian]) lt.Module[util_word.BigEndian] {
 	//
 	var (
```
DavePearce, line 132: This function is getting out of hand.  It was already pretty bad, and its much worse.  Can we try and split this into specific subfunctions to give a better high-level view when looking at the function?

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -127,18 +130,23 @@ func (p *FullObserver[W, I, M]) traceModule(m machine.Module, states []State[W],
...
+		cols               []array.MutArray[util_word.BigEndian]
+		nrows              = uint(len(states))
+		multiLine          = isMultiLineFunction(m)
+		isAccessOnceMemory = isAccessOnceMemory[W](m)
+		extra              = extraColumnsForAccessOnceMemory[W](m)
 	)
 	// Initialise columns
-	if multiLine {
+	switch {
+	case multiLine:
```
DavePearce: Maybe call it `multiLineFunction` so we can see its different from a "memory" ?

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -346,31 +454,82 @@ func traceColumns[W any](regs []register.Register, cols []array.MutArray[W]) []l
...
+			states = append(states, NewState(0, false, rom.Width(), words))
+			addressUint64++
+		}
 	}
 	//
 	return states
 }
+
+// split address takes an address uint64 and returns this address split according
+// to
```
DavePearce: `according to` ... ?

auxiliary wom columns tracing.md
```diff
@@ -0,0 +1,266 @@
+# Auxiliary WOM columns tracing
```
DavePearce: I do not think this file was intended to be part of this PR.  Certainly, it is in the wrong place.  But, also, I'm sure if its that helpful.  However, if it could be significantly compacted then it could go into the file `docs/ZKC_ARCHITECTURE_AIR.md` to provide a high-level overview of constraints / tracing for memory modules.

ram.md
```diff
@@ -0,0 +1,161 @@
+# RAM module
```
DavePearce: RAM is not even covered in this issue, so I definitely don't think this file belongs as part of this PR.

rom (read once memory).md
```diff
@@ -0,0 +1,70 @@
+# ROM (read ONCE memory) module
```
DavePearce: Again, this doesn't belong here.  But, some kind of high-level summary could go into `docs/ZKC_ARCHITECTURE_AIR.md`

rom (read only memory).md
```diff
@@ -0,0 +1,94 @@
+# ROM (read ONLY memory) module
+
```
DavePearce: Again, this doesn't belong here.  But, some kind of high-level summary could go into `docs/ZKC_ARCHITECTURE_AIR.md`

wom.md
```diff
@@ -0,0 +1,130 @@
+## WOM module
```
DavePearce: Again, this doesn't belong here.  But, some kind of high-level summary could go into `docs/ZKC_ARCHITECTURE_AIR.md`

DavePearce (changes requested):
Overall, I think this is looking pretty good.  I have requested some changes, mostly related to tidying up long functions and spurious `md` docs which are in the wrong place / not really part of the PR.  **I think the only serious issue I found** is with regard to static reference tables which are treated in `FullObserver` like other ROM/WOM modules.  This doesn't make sense to me as these should have special treatment.

pkg/test/zkc_unit_test.go
DavePearce: I dug into this.  The reason is that we cannot generate invalid traces, so we cannot run these tests.  This is a known issue which we plan to address (see #1567).

pkg/test/util/check_valid.go
OlivierBBB: That was Claude's work, making stuff compile

pkg/zkc/constraints/translator.go
OlivierBBB: I used go notation for the slice of address columns. The whole address slice should vanish in padding.

pkg/zkc/vm/internal/memory/geometry.go
OlivierBBB: I'm actually a little confused whether to use line or lane. In geometry.go you can find address **lines**
```go
// AddressLines returns the number of address lines
func (p Geometry[W]) AddressLines() uint {
	return p.numInputs
}
```
But for some reason I've mostly used **lanes**. I'll switch to address lines.

pkg/zkc/vm/internal/memory/wom.go
OlivierBBB: Done ✅

pkg/zkc/compiler/codegen/statement.go
OlivierBBB: Git blame tells me it's from a commit of @letypequividelespoubelles dating back to June 24

pkg/zkc/vm/internal/trace/full_observer.go
OlivierBBB: Do we not want to fill addresses and data for static tables ? They ought to be the target of standard (non log bus) lookups, are you suggesting the prover team own these tables ?

pkg/zkc/vm/internal/trace/full_observer.go
OlivierBBB: Claude simplified the function by breaking it into smaller pieces.

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -53,10 +53,13 @@ func (p *FullObserver[W, I, M]) Initialise(machine M) {
 			// Static reference tables (e.g. the $range_un range-check tables) are
 			// not populated by execution, so their full contents must be seeded
 			// here; otherwise lookups into them would see an empty target.
-			p.trace[i] = initialiseROM(&m.ReadOnly)
+			p.trace[i] = initializeMemoryAddressesAndContents(&m.ReadOnly)
```
DavePearce, line 56: So, I think we should just drop this.

DavePearce (approved):
Overall, this looks much improved.  The only real issue is that of the static ROM, and I would suggest just commenting that out.  I have approved it, so you can merge regardless if you want.

PR #1839, opened 2026-06-03
fix: update lfdt-lineth-org runners for lint and test workflows

eloi010:
Updates GitHub Actions workflows to run CI jobs on the `gha-lfdt-lineth-ss-ubuntu-24-amd64-large` runner label, aligning test and lint execution with the intended LFDT Lineth runner scale set.


waskow-consensys (approved):
🚀

PR #1840, opened 2026-06-03
perf: replace slice bits by low bits

gbotrel:
On the hot path of `zkc exec` , micro bench on this specific function x5 - x10, on some specific `zkc exec` runs, x2.

### DIFF
--- a/pkg/zkc/vm/internal/word/uint.go
+++ b/pkg/zkc/vm/internal/word/uint.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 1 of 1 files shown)


DavePearce (approved):
Overall, LGTM --- it is definitely an improvement.  However, I have to note that on two somewhat realistic benchmarks I am not observing much performance benefit.  I guess its not hot.  The benchmarks are: 50K block keccak (ZkC) which takes around `58s` vs `56s`; a run of blake on the RISC-V interpreter (not sure how many rounds, but takes around `4s` for both versions).

PR #1842, opened 2026-06-07
[RV] `#[cost:<identifier>]` annotation proposal

### LINKED ISSUE #1841: feat: cost annotation proposal
Add an optional annotation on top of lines of a zkc program, so as to count the number of zkc internal representation instructions that are generated. This is specifically useful to estimate the cost of decoding. Here is an example:

```
#[cost:i_type_decoding]
  imm12::rs1::funct3::rd = instruction_parameters
```

lorenzogentile404:


DavePearce (changes requested):
This PR doesn't make sense in allowing annotations on statements.  Doing this adds significant complexity for very limited benefit.  The key benefit would come from reporting instruction counts on a per-function basis.  Specifically, we do not actually need an annotation even for this.  Rather, we want some kind of `--profiling` option on the `debug` command than then reports the dynamic costs on a function-by-function basis.

My suggestion is to start over with this new objective: report the number of instructions executed by each function when `zkc debug --profiling` is given.  Initially, it is fine to report just the cumulative cost for each function (i.e. it includes that of the functions it calls).  Eventually, we might want to separate this out --- but its a separate PR.

PR #1846, opened 2026-06-09
feat: adds support for more opcodes in the bytecode interpreter 

gbotrel:
Adds support for more opcodes in the bytecode interpreter, and more test coverage in the zkc basic unit tests.


pkg/zkc/vm/internal/bytecode/arith.go
```diff
@@ -69,10 +69,28 @@ func (p *Arith[W]) Codes(_ uint32) []uint32 {
...
 	case n == 1 && m == 1 && cz:
 		return encodeMove_1s1(p.Source[0], p.Target[0])
-	case n == 2 && m == 1 && cz:
-		return encodeArith_2n1(p.Op, p.Source[0], p.Source[1], p.Target[0])
+	case n == 1 && m == 1:
+		return encodeArith_1n1c(p.Op, p.Source[0], p.Target[0], p.Constant)
+	case n == 2 && m == 1:
+		// x*y*0 is always 0: fold to a constant load so the intermediate
+		// product x*y cannot raise a spurious overflow error.
+		if p.Op == arithop_MUL && p.Constant.Cmp64(0) == 0 {
```
DavePearce, line 77: This is a compiler optimisation.  I don't think we should get into this, as it opens a whole can of worms.  We can just assume that this kind of optimisation is applied before we get here.

pkg/zkc/vm/internal/bytecode/arith.go
```diff
@@ -69,10 +69,28 @@ func (p *Arith[W]) Codes(_ uint32) []uint32 {
...
+			return encodeLdc_1(p.Constant, p.Target[0])
+		}
+		// There is no 2-source-plus-constant instruction form, so compute
+		// "x op y" first, then fold in the constant (when used) with a second
+		// one-source instruction operating in place on the target.
+		codes := encodeArith_2n1(p.Op, p.Source[0], p.Source[1], p.Target[0])
```
DavePearce, line 83: Ok, maybe put this into its own function `encodeArith_2n1c`.  This kind of transformation is fine for now, but it does introduce problems.  Specifically, we cannot reverse the compiled bytecode and get back the original bytecode.

pkg/zkc/vm/internal/bytecode/arith.go
```diff
@@ -99,6 +117,13 @@ func decodeArith[W word.Word[W]](pc uint32, codes []uint32) (Bytecode[W], uint32
 		targets = []Reg{rd}
 		constant = word.Const64[W](1)
 		op = opcodeToArithOp(opcode)
+	case ADDC, SUBC, MULC:
```
DavePearce, line 120: Maybe set the constant names accordingly to e.g. `ADDC_1n1` or perhaps `ADD_1n1c`.

pkg/zkc/vm/internal/bytecode/bytecode.go
```diff
@@ -103,6 +103,8 @@ const (
 	SUBC
 	// MULC (multiply with constant) instruction
 	MULC
+	// ARITHV (arithmetic with vector target) instruction
+	ARITHV
```
DavePearce, line 107: This does not make sense to me.  I need to have a look at what's going on.  But, I guess it means it will do a "double" decoding during execution, which is the intended approach.

pkg/zkc/vm/internal/bytecode/call.go
```diff
@@ -0,0 +1,114 @@
...
+	return &Call{id, args, returns}, n
+}
+
+// ============================================================================
+// CALL instruction. Format of this instruction is:
+//
+//	31                                0
+//
+// +--------+--------+--------+--------+
+// | nrets  | nargs  |   id   | opcode |
```
DavePearce, line 71: Yeah, this is not the design I intended.  We don't want if.  We want a physical address.

pkg/zkc/vm/internal/bytecode/cat.go
```diff
@@ -0,0 +1,99 @@
...
+// +--------+--------+--------+--------+
+// | tgt3   | tgt2   | tgt1   | tgt0   |
+// +--------+--------+--------+--------+
+// | ... packed source registers ...    |
+// +------------------------------------+
+//
+// The first source and target are the least-significant limbs.
+// ============================================================================
+
+func encodeCat(targets []Reg, sources []Reg) []uint32 {
```
DavePearce, line 70: eventually we will want some specialisations here.

pkg/zkc/vm/internal/bytecode/interpreter.go
```diff
@@ -60,8 +68,8 @@ type Interpreter[W word.Word[W]] struct {
 	// Data stack holding the activation records (registers) of active function
 	// calls.  The current frame begins at fp.
 	dataStack heap.Heap[W]
-	// Call stack holding return addresses (bytecode offsets) for nested calls.
-	callStack heap.Heap[uint32]
+	// Call stack holding caller state for nested calls.
+	callStack []callFrame
```
DavePearce, line 72: I think this is a mistake.  Most likely, I will undo some of this if we merge the PR, and continue along the original design.

DavePearce: The call frame should be just the `fp` and the return address.

DavePearce (dismissed):

lorenzogentile404 (approved):

PR #1851, opened 2026-06-11
feat: refactor bipartite => paged memory

### LINKED ISSUE #1850: feat: refactor bipartite RAM to paged RAM
## Summary

Under the current model, the `#[bipartite]` annotation is used to indicate a "bipartite" memory.  However, this is no longer required.  Instead, we want an annotation [and implementation] of _paged_ memory.  That is, where memory is divided into chunks of `n` words (e.g. `1M` pages, etc).

The reason for supporting a paged format is to twofold: firstly, we might expect such memory is not used entirely and, hence, paging reduces overall memory footprint; secondly, it enables the computation of "memory delta's" (see below)

## Notes

A key requirement of the system is the ability to generate "memory checkpoints" efficiently.  For example, we might want to generate a memory checkpoint every 1000 times a given function is executed.  One obvious option here is to clone the entire memory at each point, but this could be prohibitively expensive.  Instead, we can optimise this cloning process through paging with e.g. a "dirty" bit (i.e. which indicates the given page was assigned in the previous segment).  This is essentially a copy-on-write scheme, where by the first write to a page where `!dirty` causes a clone and sets `dirty`.  Thus, subsequent writes do not cause a clone.  This imposes an additional overhead on every write to a paged memory, but should reduce cloning cost provided that most writes are not dirty (i.e. occur across a small set of pages).

Finally, it is worth noting that cloning of standard (i.e. non-paged) memory is always done in its entirety.  The assumption here is that such memory is relatively small, and hence cloning is efficient and compact.

DavePearce:
This refactors the existing notion of a "bipartite memory" into a new concept called "paged memory".  This has a different usage model which assumes high memory addresses are not accessed.  Instead, the intuition behind paged memory is that it supports efficient checkpointing for memories which are assumed to be (potentially) quite large.


pkg/zkc/vm/internal/memory/paged_ram.go
```diff
@@ -0,0 +1,246 @@
...
+
+	"github.com/LFDT-Lineth/zkc/pkg/schema/register"
+	"github.com/LFDT-Lineth/zkc/pkg/trace"
+	"github.com/LFDT-Lineth/zkc/pkg/util"
+)
+
+// PAGE_SIZE determines the number of words in a single page of a
+// PagedRandomAccess memory.  This is fixed at 1M words.
+const PAGE_SIZE uint64 = 1 << 20
+
```
lorenzogentile404, line 27: Should we make this configurable to make easier to test performances with different values_

lorenzogentile404 (dismissed):
LGTM. Left a comment.

lorenzogentile404 (approved):

PR #1853, opened 2026-06-11
feat: support #[debug] annotation on functions

### LINKED ISSUE #1852: feat: support `#[debug]` function annotation
## Summary

The purpose of this annotation is to enable a function to be marked as "debug only" and, hence, omitted when running in non-debug mode.

DavePearce:
This adds support for a #[debug] annotation on functions.  This means that calls to functions with this annotation are elided when running with "--quiet".  Specifically, it means there is no overhead resulting from these functions.  Such functions cannot return values and should not modify memory (for obvious reasons).

A current restriction is that debug functions cannot write to memories (either wom or ram).  Furthermore, they cannot call non-debug functions. Eventually, if a use case for this was required, we could permit them to call "pure" functions or, alternatively, introduce a notion of a read-only memory effect.


lorenzogentile404 (approved):
LGTM. However, what I had in mind was to create 2 versions of `printf`, for example `printf` and `log`. The behaviour is the same, except that `printf` always prints its content, while `log` only if we are in verbose mode or so.
We probably can achieve an equivalent behaviour with this solution. Is this solution you implemented simpler from a `zkc` perspective?

testdata/zkc/unit/debug_02.zkc
```diff
@@ -0,0 +1,23 @@
...
+    if (lhs + rhs) != res {
+        fail
+    }
+    return
+}
+
+#[debug]
+fn crash(x:u8) {
+    printf "x = %d\n", x
+    fail
```
lorenzogentile404, line 22: This is very much a corner case, but since execution of debug functions should not affect execution, maybe we should not allow to have a `fail` inside.

PR #1860, opened 2026-06-11
feat: gogen — compile ZkC word machines to native Go

gbotrel:
Native-Go fast-execution tier, generated from `WordMachine[Uint]` (either shape, unsplit). Fast mode only — no checkpoints, no tracing.

**keccak-f ×50,000, plain shape (M5 Max):**

| executor | time |
|---|---|
| gogen | **0.22 s** |
| bytecode | 15.4 s |
| wordmachine-u64 | 24.3 s |

Lowered shape ×500: gogen 0.79 s (incl. 177 ms `go build`) vs Uint reference 83.8 s. The u64-narrowed tiers cannot run that shape (u65 temporaries).

**Additions**
- `pkg/zkc/vm/internal/gogen` (~2.1 KLOC, ≈ bytecode pkg size): per-width type lowering (u64 / two-limb ≤128), interval analysis deletes provably-dead store checks (`NoIntervals` opt-out), single panic/recover error path, SROM/ROM/WOM/RAM/paged memory roles.
- `pkg/zkc/gogen`: sha256-keyed single-flight build cache (persistent dir = reuse across processes) + JSON run protocol.
- `zkc generate [-o file.go] [--pkg name] file.zkc…`: AOT artifact; header carries the source digest for `go:generate` staleness checks.

**Verification** — oracle is always the Uint reference executor: every `Test_ZkcUnit` fixture cross-checked (outputs + verdicts, zero silent skips), golden files pin the emitted shape, opcode/condition enum walk breaks the build on upstream additions, full-corpus generation sweep, ≥10⁴-vector fuzz: zero divergence. Generated code is `go vet`-clean.


pkg/zkc/compiler/validate/entry_point.go
```diff
@@ -0,0 +1,38 @@
...
+// any declared parameter would silently read as zero in every execution mode
+// (reference interpreter, bytecode and generated Go alike).  Program inputs
+// must instead flow through declared input memories, which are initialised
+// from the supplied inputs at boot time.  Hence, an entry point declaring
+// parameters is almost certainly a mistake, and is rejected here.
+func EntryPoint(program ast.Program, srcmaps source.Maps[any]) []source.SyntaxError {
+	var errors []source.SyntaxError
+	//
+	for _, d := range program.Components() {
+		if fn, ok := d.(*decl.ResolvedFunction); ok && fn.Name() == "main" && fn.NumInputs > 0 {
```
DavePearce, line 31: At some point, I would prefer not to have `main` hardcoded here.  Instead, we should somewhere have a parameter stating the name of the boot function.  This is needed, for example, to prune unreachable code.

pkg/zkc/gogen/cache.go
```diff
@@ -0,0 +1,140 @@
...
+type build struct {
+	once sync.Once
+	path string
+	err  error
+}
+
+// NewCache returns a cache rooted at dir.  An empty dir means a temporary
+// directory created on first use (per-process cache); a fixed dir makes the
+// cache persistent — binaries found there are reused without rebuilding.
+func NewCache(dir string) *Cache {
```
DavePearce, line 59: This does not seem to be used.

pkg/zkc/gogen/run.go
```diff
@@ -0,0 +1,63 @@
...
+	return RunRaw(prog, inJSON)
+}
+
+// RunRaw is Run with the inputs already marshalled to JSON — benchmarks
+// pre-marshal once so the timed loop measures the executor, not
+// json.Marshal of the inputs.
+//
+// The protocol matches the generated main harness: JSON inputs on stdin,
+// JSON outputs on stdout, exit code 1 for an execution error.
+func RunRaw(prog string, inJSON []byte) (map[string][]uint64, bool, error) {
```
DavePearce, line 41: This seems like it should go somewhere like `vm/internal/gogen/util` to support the testing framework.

pkg/zkc/vm/internal/gogen/testdata/golden/branch.go.golden
```diff
@@ -0,0 +1,100 @@
+// Code generated by zkc gogen. DO NOT EDIT.
```
DavePearce, line 1: So, these are what ... regression tests?  I'm kind of unsure whether we need this, since the testing framework covers functional correctness of the generated code.

DavePearce (dismissed):
Overall, LGTM and can be merged.  I have some suggested changes, and some other things (which could separate PRs) and also left some comments.  There are comments from Cursor which could be addressed.  

**Also**, tbh the code in `zkc/vm/internal/gogen/` is ... "icky".  This will be hard to work with, and I imagine we will want to clean it up at some point.  But, this is a prototype and its doing its job.

## Suggestions

* `memGrow`.  Can we rename this to `memWrite` please?
* `zkc generate -o`.  Seems to fail as it puts `package test` at the top
* **Bug**.  I found this:
```
fn add(x:u64, y:u64) -> (r:u64) {
   var tmp:u64
   tmp::r = (x as u128) + (y as u128) + 1
}
```

This causes a problem with the type of the constant `1`.

## Others

* **Input Format**.  The input format accepted by the generated `go` file differs from that accepted by e.g. `zkc exec`.  Specifically, `zkc exec` accepts a json with a field for each input that is a hex string.  Being consistent would help I think.
* `printf` / `fail`.  Currently, `printf` does nothing whilst `fail` does not include any messages.  It would be good to support this at some point.

pkg/zkc/gogen/cache_test.go
```diff
@@ -0,0 +1,106 @@
...
+	)
+
+	for i := range paths {
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
+
+			paths[i], errs[i] = c.Build(trivialMain)
+		}()
```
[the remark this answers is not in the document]
DavePearce: Variable capture is annoying in cases like this!

pkg/test/zkc_exec_micro_bench_test.go
```diff
@@ -0,0 +1,226 @@
...
+			wm := compileMicro(b, tc.src)
+			inputs := map[string][]vm.Uint{"data": {uintWord(microSteps)}}
+
+			b.ResetTimer()
+
+			for i := 0; i < b.N; i++ {
+				keccakByteSink = runKeccakCore(b, compileMicro(b, tc.src), inputs)
+				_ = wm
+			}
+		})
```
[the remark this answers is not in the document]
DavePearce: I guess this is not an issue because the code calls `ResetTimer`.

pkg/zkc/vm/internal/gogen/gen.go
```diff
@@ -0,0 +1,1009 @@
...
+		regs     = fn.Registers()
+		outNames []string
+		retType  string
+		paramFmt []string
+		params   = uint(0)
+	)
+
+	if !isBoot {
+		params = ni
+		for i := range ni {
```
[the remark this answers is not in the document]
gbotrel: This is valid since Go 1.22 — `range` over any integer type (including `uint`) iterates 0..n-1. The package compiles and is exercised by the test suite.

pkg/zkc/vm/internal/gogen/gen.go
```diff
@@ -0,0 +1,1009 @@
...
+			}
+
+			if l.width > 64 {
+				paramFmt = append(paramFmt, fmt.Sprintf("%s, %s uint64", l.lo(), l.hiName()))
+			} else {
+				paramFmt = append(paramFmt, fmt.Sprintf("%s uint64", l.lo()))
+			}
+		}
+
+		for i := range no {
```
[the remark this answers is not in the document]
gbotrel: Same as above — `range` over an integer is valid Go since 1.22.

pkg/zkc/gogen/cache_test.go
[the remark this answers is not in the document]
gbotrel: Not an issue since Go 1.22: loop variables are per-iteration, so each goroutine captures its own `i`. (The module requires go 1.24.)

pkg/cmd/zkc/generate.go
```diff
@@ -0,0 +1,149 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package zkc
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"path"
+	"strings"
```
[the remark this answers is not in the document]
gbotrel: Resolved in f01ea4b7 by removing the filename derivation altogether: the package now defaults to `main` (runnable harness) and `--pkg` names the importable case explicitly, so the `path` import is gone.

pkg/cmd/zkc/generate.go
```diff
@@ -0,0 +1,149 @@
...
+func packageName(pkg, output string) string {
+	if pkg != "" {
+		return pkg
+	}
+
+	if output == "" {
+		return "main"
+	}
+	// Derive a valid Go identifier from the output file's base name.
+	base := strings.TrimSuffix(path.Base(output), ".go")
```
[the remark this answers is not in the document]
gbotrel: Moot after f01ea4b7 — the package name is no longer derived from the output filename.

pkg/test/zkc_exec_micro_bench_test.go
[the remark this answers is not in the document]
gbotrel: Good catch — `ResetTimer` didn't actually cover this since `compileMicro` ran *inside* the timed loop. Fixed in f01ea4b7: the machine is compiled once and reused across iterations (`Boot` reinitializes state, same as the keccak bench).

pkg/zkc/compiler/validate/entry_point.go
gbotrel: Agreed — filed #1869 to parameterize the boot-function name (it's also hardcoded in the VM boot path and the generator's boot-frame detection; one config point would serve all three and enable unreachable-code pruning).

pkg/zkc/gogen/cache.go
gbotrel: It is used: `defaultCache` (backing the package-level `Build`) is `NewCache("")`, and the tests construct caches over fixed directories. It's exported for the persistent case — a prover rooting the cache at a fixed dir so binaries survive restarts.

pkg/zkc/gogen/run.go
gbotrel: It can't live under `vm/internal/...`: the consumers are `pkg/test/util/check_gogen.go` and the exec benchmarks in `pkg/test`, which couldn't import it from there (internal visibility is rooted at `pkg/zkc/vm`). Keeping it in `pkg/zkc/gogen` next to `Build`, which is the public build+run surface.

pkg/zkc/vm/internal/gogen/testdata/golden/branch.go.golden
gbotrel: Yes, regression pins on the emitted shape. The differential tests prove the code computes the right thing, but not that it stays *good* — a regression that, say, stops eliding dead width-checks or reverts single-limb registers to pairs would still pass them. The goldens make that visible in review (`go test -run TestGolden -update` regenerates). Keeping them at least for now, cheap to drop later.

gbotrel:
Thanks for the review @DavePearce — addressed in f01ea4b7:

- **Constant-typing bug**: confirmed and fixed. The `tmp::r = (x as u128) + (y as u128) + 1` repro generated `lo, hi := 1, uint64(0)`, typing `lo` as `int` and breaking the `bits.Add64` chain. Emitted constants now carry an explicit `uint64(...)` conversion, and the repro is pinned as the `wideConstAdd` differential case (which builds and runs the generated code against the reference executor).
- **`memGrow` → `memWrite`**: renamed (goldens regenerated).
- **`zkc generate -o`**: fixed — the package now defaults to `main` (with the standalone harness) regardless of `-o`; deriving the name from the filename was the source of the unrunnable `package test`. The importable case is now always explicit via `--pkg`.
- **Micro-benchmark**: Copilot's point was real after all — the compile was *inside* the timed loop, past `ResetTimer`; it now reuses the compiled machine.

Follow-ups filed for the rest:
- #1867 — align the generated harness input format with `zkc exec` (hex-string JSON)
- #1868 — `printf` output and `fail` messages in generated code
- #1869 — parameterize the entry-point name instead of hardcoding `main`

And agreed on the "icky" — the emitter layer is due a structural cleanup once the shape settles; happy to take that as a separate PR.

pkg/zkc/gogen/cache_test.go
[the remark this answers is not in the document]
DavePearce: Huh, is that right --- good to know.

pkg/zkc/compiler/validate/entry_point.go
DavePearce: Yeah, that is essentially the plan.  There will be one point where it exists and from that we will prune unreachable code.

pkg/zkc/gogen/cache.go
DavePearce: Ok, I do see that.  Its really only used for testing though.

pkg/zkc/vm/internal/gogen/testdata/golden/branch.go.golden
DavePearce: Fair enough

DavePearce (approved):
LGTM

PR #1865, opened 2026-06-12
feat: support function inlining in zkc

### LINKED ISSUE #1632: feat: support function inlining in ZkC
_(see also #1630)_

# Summary

The goal of this feature is to support function inlining in ZkC directed by an `@inline` attribute.  

## Minimal Example

```
fn f(x:u32) -> (y:u32) {
   y = 2 * g(x)
}

#[inline]
fn g(x:u32) -> (y:u32) {
   y = x + 1
}
```

Then, the compiler is responsible for automatically converting the above into the following form:

```
n f(x:u32) -> (y:u32) {
   y = 2 * (x + 1)
}
```

As usual, the goal of this optimisation is to reduce the overhead of function's.  In ZkC these can be more expensive than in a generate compute setting, since functions correspond with modules in the final AIR constraints.

DavePearce:


letypequividelespoubelles (approved):
LGTM

PR #1866, opened 2026-06-12
perf: optimize power of two div / remainder in fast mode

### LINKED ISSUE #1861: perf: for fast mode, turn / and % of power of two into << and &

letypequividelespoubelles:


letypequividelespoubelles:
It has 0 impact, thus closing:
with optim: https://github.com/LFDT-Lineth/lineth-monorepo/actions/runs/27539415795
without optim: https://github.com/LFDT-Lineth/lineth-monorepo/actions/runs/27539415795

letypequividelespoubelles:
finally a tiny win: https://github.com/LFDT-Lineth/lineth-monorepo/actions/runs/27591513920

DavePearce (approved):
Ok, looks fine.  I think if we had constants available as operands (rather than just registers) it would be a lot less "ugly" as you say.  The problem is knowing that a register holds a constant value.  So, when we fix the constant issue then we can come back and update this.

PR #1873, opened 2026-06-15
feat: support skiptable instruction

### LINKED ISSUE #1871: feat: `vm.MultiSkip` instruction
## Summary

The purpose of this feature is to add support for a "multi-way" skip function.  This roughly would correspond with a jump table.  Specifically, it contains a list `(value,skip)` pairings.

DavePearce:


letypequividelespoubelles (approved):
LGTM, but imho, the skip_multi should be plugged, as the skip_if to the guard factorization (ie generate boolean condition).

PR #1875, opened 2026-06-15
feat: factorize at wir level

### LINKED ISSUE #1793: feat: optimise constraints arising from control-flow
## Summary

Consider the following simple program (`basic_32.zkc`):

```
pub input data(address:u16) -> (byte:u16)

fn main() {
    var n:u16 = data[0]
    var r:u16 = data[1]
    var i:u16 = 0
    var v:u16 = 1
    //
    while i != n {
        v = v + v
        i = i + 1
    }
    //
    if v != r {
        fail
    }
}
```

This generates the following MIR constraints:

```
```

And then the following (abbreviated) AIR constraints:

```
(module data)

(inputs
   (address u16 0x0)
)

(outputs
   (byte u16 0x0)
)


(module main)

(computed
   (n u16 0x0)
   (r u16 0x0)
   (i u16 0x0)
   (v u16 0x0)
   ($4 u0 0x0)
   ($5 u1 0x0)
   ($pc u2 0x0)
   ($ret u1 0x0)
   ("(inv main:$pc)" 𝔽 0x0)
   ("(inv (- main:$pc 1))" 𝔽 0x7f000000)
   ("(inv (- main:n main:i))" 𝔽 0x0)
   ("(inv (- main:r main:v))" 𝔽 0x0)
   ("(inv (- main:$pc 2))" 𝔽 0x3f800000)
   ("(inv (- (shift main:n -1) main:i))" 𝔽 0x0)
   ("(inv (- (shift main:r -1) main:v))" 𝔽 0x0)
)
...
(vanish main:pc0#0 (* (- 1 (* (- $pc 1) "(inv (- main:$pc 1))")) $4))
(vanish main:pc0#1 (* (- 1 (* (- $pc 1) "(inv (- main:$pc 1))")) (- $5 1)))
(vanish main:pc0#2 (* (- 1 (* (- $pc 1) "(inv (- main:$pc 1))")) i))
(vanish main:pc0#3 (* (- 1 (* (- $pc 1) "(inv (- main:$pc 1))")) (- v 1)))
(vanish main:pc0#4 (
   *
   (- 1 (* (- $pc 1) "(inv (- main:$pc 1))"))
   (- n i)
   (- (shift $pc "1") 2)))

(vanish main:pc0#5 (* (- 1 (* (- $pc 1) "(inv (- main:$pc 1))")) (- n i) $ret))
(vanish main:pc0#6 (
   *
   (- 1 (* (- $pc 1) "(inv (- main:$pc 1))"))
   (- 1 (* (- n i) "(inv (- main:n main:i))"))
   (- r v)))

(vanish main:pc0#7 (
   *
   (- 1 (* (- $pc 1) "(inv (- main:$pc 1))"))
   (- 1 (* (- n i) "(inv (- main:n main:i))"))
   (- 1 (* (- r v) "(inv (- main:r main:v))"))
   (- $ret 1)))
...
```

What we see is a lot of repetition in the generated constraints (e.g. ` (- 1 (* (- $pc 1) "(inv (- main:$pc 1))"))`)

## Approach

A simple approach to this optimisation is to transform a ZkC program into a form where conditionals are represented using binary "control lines".  So, for example, this:

```
fn max(x:u16, y:u16) -> (z:u16) {
   if x > y {
      z = x
   } else { z = y }
}
```

becomes this:

```
fn max(x:u16, y:u16) -> (z:u16) {
   var $1:u1 = x > y
   //
   if $1 {
      z = x
   } else { z = y }
}
```

There are some challenges with this, as care must be taken to ensure the transformation produces an equivalent program.

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/factor_skip_conditions.go
```diff
@@ -0,0 +1,182 @@
...
+		alloc = register.NewAllocator[int](fn.RegisterMap())
+	)
+
+	for i, insn := range code {
+		// Decide up-front which SkipIf codes in this vector are worth factoring.
+		// This needs the whole vector body (to size each branch), which the Map
+		// closure cannot see one instruction at a time.
+		factor := factorableSkips(insn.Codes, alloc)
+		//
+		ncode[i] = insn.Map(func(idx uint, ith WordInstruction) []WordInstruction {
```
DavePearce, line 69: Great to see it is using `insn.Map()` !!

DavePearce (approved):
Ok, overall I am happy with what you've done here and how its implemented (e.g. using `insn.Map()` works well).  However, I am not happy with the actual end result.  Its not making the constraints better, as far as I can tell.  There are some reasons for this, such as missing constants in `skip_if`.  Also, the branch table optimiser is not cancelling constraints very well.  For example, looking at the MIR output for this example:

```
fn max(x:u8, y:u8) -> (z:u8) {
   //
   if x == y {
      z = x
   } else if x == 0 {
      z = 0
   } else  { z = y }
}
```

I see this:

```
(vanish max:pc0
   (∧ (== $5 0)
    ...
      (if
         (∨
            (∧ (== x $3) (!= $4 $5) (!= $6 $7))
            (∧ (!= x $3) (!= $4 $5) (!= $6 $7)))
         (== z y))))
```

My feeling is that this this should cancel from `(if ... (== x y))` to just `(== x y)`.  But, unfortunately, the cancelling tool is failing to manage it.

So, I have approved this PR since it does what I wanted.  But, I think we still have work to do to actually make this a net positive.  At the moment, it feels like it should be disabled by default.

PR #1882, opened 2026-06-15
chore: refacto fastMode flag

### LINKED ISSUE #1880: refacto compilation flag names

letypequividelespoubelles:


amkCha (dismissed):

pkg/cmd/zkc/root.go
```diff
@@ -143,7 +144,7 @@ func init() {
 	rootCmd.PersistentFlags().Bool("fir", false, "Output Field-level Intermediate Representation (FIR)")
 	rootCmd.PersistentFlags().Bool("mir", false, "Output Mid-Level Intermediate Representation (MIR)")
 	rootCmd.PersistentFlags().Bool("air", false, "Output Arithmetic Intermediate Representation (AIR)")
-	rootCmd.PersistentFlags().Bool("lower-native", false, "Lower ZkC native functions into arithmetic instructions")
+	rootCmd.PersistentFlags().Bool("fastMode", false, "Fast-mode execution (harvest partial trace information)")
```
DavePearce: No, please.  Not `--fastMode`.  Given I am going to type this like a billion times, I think `--fast` would be more appropriate with a `-f` abbreviation.

DavePearce: I would also not say this: "harvest partial trace information"

When run from the command-line like this, its not going to anything other than execute as fast as possible.  Maybe just say "(no tracing)" or something.

pkg/zkc/compiler/codegen/config.go
```diff
@@ -18,12 +18,12 @@ import "github.com/LFDT-Lineth/zkc/pkg/util/field"
...
 // setters below.
 var DEFAULT_CONFIG = Config{
-	field:          field.KOALABEAR_16,
-	inlining:       true,
-	lowerZkcNative: false,
-	quiet:          false,
-	vectorize:      true,
-	splitting:      false,
+	field:     field.KOALABEAR_16,
+	fastMode:  false,
```
DavePearce, line 22: The term `fastMode` does not make sense here.  This is a compilation step, not an execution step.  Call it what it is: `lowering`.

pkg/cmd/zkc/root.go
```diff
@@ -97,7 +97,8 @@ func runFieldAgnosticCmd(cmd *cobra.Command, args []string, cmds []FieldAgnostic
 func GetBuildConfig[F field.Element[F]](cmd *cobra.Command, field field.Config) BuildConfig[F] {
 	var build BuildConfig[F]
 
-	lowerNative := GetFlag(cmd, "lower-native") || GetFlag(cmd, "mir") || GetFlag(cmd, "air")
+	mirOrLower := GetFlag(cmd, "mir") || GetFlag(cmd, "air")
```
DavePearce: Maybe `mirOrAir`?  Confused me otherwise.

DavePearce (changes requested):
Ok, basically looks good.  Requesting some minor changes.

pkg/zkc/compiler/codegen/config.go
letypequividelespoubelles: well, the flag name should be the same everywhere, I don't want to have different naming in config.go and elsewhere ...
PLus you're compiling zkc to get the fast executor ... and it does more stuff than only lowering.

pkg/cmd/zkc/root.go
```diff
@@ -97,7 +97,8 @@ func runFieldAgnosticCmd(cmd *cobra.Command, args []string, cmds []FieldAgnostic
 func GetBuildConfig[F field.Element[F]](cmd *cobra.Command, field field.Config) BuildConfig[F] {
 	var build BuildConfig[F]
 
-	lowerNative := GetFlag(cmd, "lower-native") || GetFlag(cmd, "mir") || GetFlag(cmd, "air")
+	mirOrAir := GetFlag(cmd, "mir") || GetFlag(cmd, "air")
+	fastMode := GetFlag(cmd, "fastMode") && !mirOrAir
```
DavePearce: Still problem -- sorry

DavePearce (approved):
LGTM

PR #1883, opened 2026-06-15
feat: align gogen harness I/O with zkc exec (#1867)

### LINKED ISSUE #1867: gogen: align generated harness input format with zkc exec
Follow-up from #1860 review (@DavePearce):

> **Input Format**. The input format accepted by the generated `go` file differs from that accepted by e.g. `zkc exec`. Specifically, `zkc exec` accepts a json with a field for each input that is a hex string. Being consistent would help I think.

Today the generated `package main` harness reads `map[string][]uint64` JSON on stdin (e.g. `{"data":[3,4]}`), while `zkc exec` takes per-input hex strings. The harness, `pkg/zkc/gogen.Run`/`RunRaw`, the test harness (`pkg/test/util/check_gogen.go`) and the exec benchmarks would all need to move together.

gbotrel:
Closes #1867.

The generated `package main` harness read `map[string][]uint64` cells; `zkc exec` reads one hex string per memory. Now they share the same JSON encoding — a file feeding `zkc exec` feeds the generated binary unchanged.

- Harness bakes in each memory's data-line widths and emits self-contained pack/unpack helpers (it can't import the VM as standalone `package main`).
- `gogen.Run`/`RunRaw` move to the byte domain, mirroring `vm.BootAndExecute`.
- Importable package (`--pkg`) keeps its cell-based `Run`; golden output unchanged.
- Field-element data lines have no fixed width (the byte format can't carry them, nor can `zkc exec`) → main harness rejects them, pointing to `--pkg`.

Output is JSON hex (`{"result":"0x…"}`) rather than `zkc exec`'s `result = 0x…` print — symmetric and parseable for the round-trip.

Tests: differential corpus (`check_gogen`), fuzz, golden, coverage, field, micro/keccak agree all pass.


pkg/test/zkc_exec_keccak_bench_test.go
```diff
@@ -33,7 +33,6 @@ package test
```
DavePearce, line 33: Why do we have this test?  It shouldn't be needed, as we already have the testing framework.  The existing test `keccakf.zkc` was updated recently and it matches `keccakf_v2.zkc`.

pkg/test/zkc_exec_micro_bench_test.go
```diff
@@ -21,7 +21,6 @@ package test
```
DavePearce, line 21: Ditto --- what is this test doing that cannot be done with existing framework?

DavePearce: Basically, it generates a report which shows timing of tests for `gogen` versus interpreter.  We should integrate this into the existing framework.

pkg/zkc/vm/internal/gogen/gen.go
```diff
@@ -471,27 +488,137 @@ func (g *generator) emitMemHelpers(c *code) {
...
+	c.line("var raw map[string]string")
+	c.line("if err := json.NewDecoder(bufio.NewReader(os.Stdin)).Decode(&raw); err != nil {")
+	c.line(`fmt.Fprintln(os.Stderr, "decode:", err)`)
+	c.line("os.Exit(2)")
+	c.line("}")
+	c.line("in := map[string][]uint64{}")
+
+	for _, m := range g.inputs {
+		c.linef("in[%q] = decodeBytes(parseInput(raw[%q]), %s)", m.name, m.name, widthsLiteral(m.dataWidths))
+	}
```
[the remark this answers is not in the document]
DavePearce: Ideally, it would be good to flag unexpected or missing inputs.

DavePearce (approved):
Overall, LGTM.  I left some comments, but these can be addressed in separate PRs.

PR #1884, opened 2026-06-15
feat(gogen): printf/fail output, ≤128-bit add/mul, include dedup

### LINKED ISSUE #1868: gogen: support printf output and fail messages in generated code
Follow-up from #1860 review (@DavePearce):

> `printf` / `fail`. Currently, `printf` does nothing whilst `fail` does not include any messages. It would be good to support this at some point.

The generator currently drops `printf` instructions and reports `fail` as a bare exit-code-1 with no message. Generated code should emit printf formatting (when not built with `--quiet`) and surface the fail message on stderr so the harness/runner can relay it.

gbotrel:
Makes the RISC-V interpreter compile and run through gogen, and closes #1868.

**Include dedup.** The same file reached via two relative spellings (`memory.zkc` vs a library's `../../riscv/memory.zkc`, which `filepath.Join` can't collapse) was parsed twice → spurious "duplicate declaration". Now keyed on the absolute path. This alone blocked compiling the interpreter.

**printf/fail (#1868).** DEBUG emits printf; FAIL surfaces `machine panic: <msg>` — both byte-for-byte identical to the interpreter. Present only in non-quiet builds (compiler strips printf under `--quiet`), so quiet codegen is unchanged. The hot path is specialised writes (`WriteString`/`strconv.AppendUint`/raw byte) to a buffered stderr, not `fmt.Fprintf`: non-quiet keccak 9.85s → 0.32s (buffer) → 0.20s (specialised). A producer/consumer goroutine was measured and rejected — ordered output ⇒ single consumer ⇒ formatting stays serial, and the channel send only adds per-item cost. Before this, any printf-on benchmark was unfair: gogen skipped the formatting work other executors did.

**128-bit guards.** add/mul rejected on the static bound (sum/product of operand maxes); relaxed to reject only when the *target* exceeds 128 bits (runtime values are ≤128-bit lo/hi pairs, width-checked at store). Unblocks the interpreter's 64×64 multiply (MULH). concat/bitwise guards are on exact widths and stay.

Tests: include-dedup regression; printf verbs/%c/wide and fail-message e2e.


pkg/zkc/compiler/compiler_test.go
```diff
@@ -0,0 +1,95 @@
+// Copyright Consensys Software Inc.
```
DavePearce: Really?  I do not like the general proliferation of tests caused from claude.

pkg/zkc/gogen/run.go
```diff
@@ -14,31 +14,47 @@ package gogen
...
-	inJSON, err := json.Marshal(in)
+// Run executes a built gogen program on the given input memories, returning
+// the output memories, whether the program reported an execution error (a
+// rejected trace — the reference machine's error path), and any harness-level
+// error (everything else).
+//
+// Inputs and outputs are raw memory bytes keyed by memory name, exactly as
+// vm.BootAndExecute consumes/produces them: the generated harness shares the
+// JSON-hex encoding `zkc exec` accepts.
+func Run(prog string, in map[string][]byte) (map[string][]byte, bool, error) {
```
DavePearce, line 32: I guess this PR is on top of the other one for input encoding / decoding?

pkg/zkc/vm/internal/gogen/field_test.go
```diff
@@ -128,9 +128,11 @@ func TestGenFieldOps(t *testing.T) {
 
 	for _, in := range vectors {
 		t.Run(inputName(in), func(t *testing.T) {
-			refOut, refErr := referenceRun(t, fieldMachine(), in)
+			inBytes := encodeInputs(fieldMachine(), in)
```
DavePearce, line 131: The tests in this file look very similar to those which are called "mixed tests".  You might find you can remove this test file entirely and just open up `gogen` to run on the mixed tests as well.

pkg/zkc/vm/internal/gogen/gen.go
```diff
@@ -189,25 +204,30 @@ func Generate(wm *machine.Word[word.Uint], cfg Config) (string, error) {
...
+	inputs        []memInfo
+	outputs       []memInfo
+	sroms         []memInfo // static read-only memories (baked contents)
+	rams          []memInfo // read-write scratch memories
+	pageds        []memInfo // paged read-write scratch memories
+	modules       []instruction.Module
+	funcByID      map[uint]*wordFunction
+	modulus       *big.Int        // the machine's prime modulus (for mod-P ops)
+	names         map[string]bool // sanitized identifiers already taken
+	usesBits      bool            // whether math/bits is referenced (decides the import)
```
DavePearce: Would be better to just use a `map[string]bool` or something to track functions should be included, rather than this ever growing list.

DavePearce (dismissed):
Also LGTM.  Some comments, which again can be addressed separately.

DavePearce (dismissed):
LGTM

DavePearce (approved):
LGTM

PR #1888, opened 2026-06-16
1864 feat support double word arithmetic instructions

### LINKED ISSUE #1864: feat: support double-word arithmetic instructions
## Summary

The goal of this feature is to enable the interpreter to support arithmetic operations with double-word precision.  The key benefit of this is that it means we can do effective `u128` arithmetic without register splitting.  An example would this case:

```
fn inc(val:u64) -> (res:u64) {
    var c:u1
    //
    c::res = (val as u65) + 1
}
```

Currently, this compiles and executes with `vm.Uint64` but the current implementation can result in an unexpected (and unnecessary) arithmetic overflow.

This feature demands a critical change in the semantics of subtraction.  Specifically, we move to a model which could be called "subtract with borrow".  For example, the following illustrates:
```
fn dec(val:u64) -> (res:u64) {
    var b:u1
    //
    b::res = (val as u65) - 1
}
```

Again, currently, this leads to an `arithmetic underflow` for e.g. `val==0x00`.  Instead of this semantic, the above should pull the "borrow" out of the register `b`.  In otherwords, `b` is set to `1` and `res` is set to `0xff`.  Whilst this is a somewhat counterintuitive semantic, it roughly mirrors what happens with addition.

DavePearce:


DavePearce:
dead on arrival

PR #1890, opened 2026-06-16
feat: support `u128` arithmetic

### LINKED ISSUE #1889: feat: support `uint128` arithmetic
This is the quickest way to get the interpreter running.

DavePearce:
Claude summary of changes introduced by this PR:

**uint128 arithmetic** (pkg/zkc/vm/internal/word/uint128.go, +502 lines)
- New Uint128{ hi, lo uint64 } type implementing the Word[Uint128] interface.
- Arithmetic implemented directly on the two words via math/bits (Add/Sub with carry, etc.), so common paths neither
allocate nor touch big.Int.
- Wired into the word-type machinery: word.go, uint.go, uint64.go, plus vm/word.go and the machine layer
(machine/base.go, machine/word.go).

**Bytecode interpreter becomes default** (pkg/zkc/vm/internal/bytecode/)
- Interpreter now the default tool for fast-mode execution (interpreter.go +94).
- Added debug output and fail output generation (debug.go, fail.go, encoder.go, new program.go).
- CLI: exec -fg can emit debug output, and exec --gogen support added (pkg/cmd/zkc/execute.go, new gogen/build.go,
moved gogen/run.go into pkg/cmd/).

**Test harness fixes** (pkg/test/util/, pkg/test/zkc_*)
- The gogen test framework was silently swallowing failing tests and recompiling the binary for every test vector;
it now compiles each .zkc file once regardless of how many vectors use it.
- New unit fixtures: testdata/zkc/unit/basic_46…48 exercising the new arithmetic.


pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+	lo, carry := bits.Add64(p.lo, w, 0)
+	hi, carry := bits.Add64(p.hi, 0, carry)
+	//
+	return Uint128{hi, lo}, carry != 0
+}
+
+// AddMod implementation for Word interface.
+func (p Uint128) AddMod(w, m Uint128) Uint128 {
+	if m.isZero() {
+		panic("modulus by zero")
```
OlivierBBB, line 55: "division by 0" maybe ? The same choice was made in `MulMod` so it seems deliberate.

pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+// Bandwidth implementation for Word interface.
+func (p Uint128) Bandwidth() uint {
+	return 128
+}
+
+// BigInt implementation for Word interface.
+func (p Uint128) BigInt() *big.Int {
+	var buf [16]byte
+	binary.BigEndian.PutUint64(buf[0:8], p.hi)
+	binary.BigEndian.PutUint64(buf[8:16], p.lo)
```
OlivierBBB, line 86: Absolutely inconsequential, but wouldn't go syntax prefer
```
	binary.BigEndian.PutUint64(buf[:8], p.hi)
	binary.BigEndian.PutUint64(buf[8:], p.lo)
```
?

DavePearce:
Cursor review

DavePearce:
bugbot run

pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+
+// MulMod implementation for Word interface.
+func (p Uint128) MulMod(w, m Uint128) Uint128 {
+	if m.isZero() {
+		panic("modulus by zero")
+	}
+	// Form the full 256-bit product, then reduce it modulo m.  There is no
+	// hardware 256/128 division, so the reduction is a (big.Int-free) bitwise
+	// long division over the product.
+	hi, lo := p.Mul(w)
```
OlivierBBB, line 164: One could (should ?) reduce `p` and `w` modulo `m` before multiplying if `w` is small (i.e. if `w.hi` is zero).

When used in conjunction with the suggested case analysis in `func (p Uint128) Mul(w Uint128) (hi, lo Uint128) { ... }` this will improve perfs.

pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+		return true
+	case bitwidth >= 64:
+		return p.hi>>(bitwidth-64) == 0
+	default:
+		return p.hi == 0 && p.lo>>bitwidth == 0
+	}
+}
+
+// Mul implementation for Word interface.
+func (p Uint128) Mul(w Uint128) (hi, lo Uint128) {
```
OlivierBBB, line 133: `Mul` carries out all potential computations but could stop early at `Word 1` or `Word 2` in certain subcases.

OlivierBBB: This implementation avoids certain multiplications

```go 
// Mul implementation for Word interface.
func (p Uint128) Mul(w Uint128) (hi, lo Uint128) {
	// Schoolbook 128x128 -> 256 bit multiply built from four 64x64 -> 128 bit
	// partial products.  Each bits.Mul64 returns its result as (high, low) word.
	r1, r0 := bits.Mul64(p.lo, w.lo) // contributes at 2^0

	// fast case: both inputs are small
	if p.hi == 0 && w.hi == 0 {
		return Uint128{0,0}, Uint128{r1, r0}
	}

	b1, b0 := bits.Mul64(p.lo, w.hi) // contributes at 2^64
	c1, c0 := bits.Mul64(p.hi, w.lo) // contributes at 2^64

	r1, k0 := bits.Add64(r1, b0, 0)
	r1, k1 := bits.Add64(r1, c0, 0)

	// Mul64 outputs (x1,x0) are bounded by
	// (2^64 - 1)^2 ≡ 11...10_00...01
	// in particular x1 is at most 2^64 - 2 so we can add 0/1 to it without overflow
	r2, k2 := bits.Add64(b1 + k0, c1 + k1, 0)

	// intermediate case: one input is large
	if p.hi == 0 || w.hi == 0 {
		// k2 ≡ 0 in this case
		return Uint128{0,r2}, Uint128{r1, r0}
	}

	// default case: both inputs are large
	d1, d0 := bits.Mul64(p.hi, w.hi) // contributes at 2^128
	r2, k3 := bits.Add64(r2, d0, 0)
	r3 := d1 + k2 + k3

	return Uint128{r3, r2}, Uint128{r1, r0}
}

```

pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+	return p.Shr64(n.lo)
+}
+
+// Shr64 implementation for Word interface.
+func (p Uint128) Shr64(n uint64) Uint128 {
+	switch {
+	case n == 0:
+		return p
+	case n < 64:
+		return Uint128{p.hi >> n, (p.lo >> n) | (p.hi << (64 - n))}
```
OlivierBBB, line 231: Absolute nitpick but
```go
return Uint128{p.hi >> n, (p.hi << (64 - n)) | (p.lo >> n)}
```
might be clearer since `p.hi << (64 - n)` is the leading bits and `p.lo >> n` the trailing bits.

pkg/zkc/vm/internal/word/uint128.go
OlivierBBB: See also `SetBigInt`, `GobEncode` and `GobDecode`

pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+		if p.hi == 0 {
+			return Uint128{0, p.lo / d}, Uint128{0, p.lo % d}
+		}
+		// 128-bit / 64-bit division.  bits.Div64 requires its high word to be
+		// strictly less than the divisor, which holds since r0 = p.hi % d < d.
+		qHi := p.hi / d
+		r0 := p.hi % d
+		qLo, r1 := bits.Div64(r0, p.lo, d)
+		//
+		return Uint128{qHi, qLo}, Uint128{0, r1}
```
OlivierBBB, line 422: LGTM

pkg/zkc/vm/internal/word/uint128.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+	// The divisor exceeds 64 bits, so the quotient is guaranteed to fit within
+	// 64 bits.  Anything smaller than the divisor divides to zero.
+	if p.Cmp(v) < 0 {
+		return Uint128{}, p
+	}
```
OlivierBBB, line 428: Shouldn't this case be the leading case ? I mean: before `v.hi == 0` ? Also the doc string looks wrong.

pkg/cmd/zkc/execute.go
```diff
@@ -63,10 +65,10 @@ func runExecuteCmd[F field.Element[F]](cmd *cobra.Command, args []string, field
 		quiet = GetFlag(cmd, "quiet")
 		// fast mode flag
 		fast = GetFlag(cmd, "fast")
-		// ultra fast mode flag
-		ultra_fast = GetFlag(cmd, "ultra-fast")
+		// gogen mode flag: execute via generated Go rather than the interpreter
+		gogen = GetFlag(cmd, "gogen")
 		// identify whether tracing required or not.
-		tracing = check || outputFile != ""
+		tracing = check || outputFile != "" || !fast
```
OlivierBBB, line 71: So `ultra-fast` and `gogen` modes do the tracing but `fast` doesn't ?

OlivierBBB (approved):
LGTM with minor comments

PR #1891, opened 2026-06-16
feat: add range constraints

### LINKED ISSUE #1721: feat: limitless type proofs for ZkC
## Summary

ZkC will require an equivalent to the "limitless" type proofs used in `zkasm` for handling large range constraints.  That means, for a type `uX`, generating an appropriate module which recursively decomposed into another (smaller) module `uY`.  Eventually, this either ends with a range constraint primitive, or a lookup in a static reference table.

## Approach

Some thoughts:

* (`vm.Instruction`) We could actually have a `vm.Instruction` (e.g. `x ∈ u16`).  The benefit of this is that, in some situations, we could optimise some checks away.
* (**Function Call**) Instead of a `vm.Instruction` could just have a function call `u16(x)` which is then traced in the usual way.  Removing duplicates would be handled as for any other function call.  As with the `vm.Instruction` approach these can be optimised away in some situations, etc.
* (**Static Reference Table**) We do not want to offload the lowest range checks to the prover, as was previously done for `zkasm`.  Instead, we want to generate static reference tables on demand.

letypequividelespoubelles:


pkg/zkc/vm/instruction/base.go
```diff
@@ -74,11 +74,38 @@ func NewDebug(chunks ...FormattedChunk) *Debug {
...
+//   - Unconditional: the lookup is always enabled, i.e. it applies to every row
+//     of the caller rather than being gated by a per-row selector line.  This is
+//     used by range-check calls, which must hold for all rows.
+//   - Static: the callee is a static table (a memory) rather than a function, so
+//     the lookup targets the table's value column and no stack frame is entered
+//     when the call is executed.
+type Call struct {
+	base.OpIo
+	// Unconditional indicates the lookup applies to all rows (no selector line).
+	Unconditional bool
```
DavePearce: I don't like the use of these two booleans `Unconditional` and `Static`:

* Firstly, I don't think we need `Unconditional`.  Basically, if its a static call then its automatically unconditional.
* Secondly, I also don't think we need `static` either.  We actually already have what we need: `MEMORY_READ` instruction, which is specifically for reading from a memory (which can be static).

The only thing for reusing `MEMORY_READ` instruction here is that it will no returns.  This may break some logic somewhere.  If the logic is hard to fix you could have a `STATIC_READ` opcode specifically for this case.  This should still fail if we attempt to read an address which doesn't exist in the target memory.

pkg/zkc/vm/internal/transform/range_constraints.go
```diff
@@ -0,0 +1,329 @@
...
+	for _, m := range modules {
+		names[m.Name()] = true
+	}
+	//
+	return names
+}
+
+// rangeModuleName returns the canonical module name for a given width.
+func rangeModuleName(w uint) string {
+	return fmt.Sprintf("range_u%d", w)
```
DavePearce: This is a mistake.  To avoid name clashes, the convention is that generated names begin with `$`.  That way, they can never clash with names at the ZkC source level.  In this case, I would recommend the name format `fmt.Sprintf("$u%d",w)`.

pkg/zkc/vm/internal/transform/range_constraints.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		// Register ids follow declaration order: value=0, lo=1, hi=2.
+		valID = register.NewId(0)
+		loID  = register.NewId(1)
+		hiID  = register.NewId(2)
+		// Destructure value into its low (lo) and high (hi) halves: value = hi::lo
+		// (little-endian targets, so lo receives the low bits).
+		destruct = instruction.UintDestruct[W](register.NewVector(loID, hiID), valID)
+		// Range-check each half via an unconditional call into its range module.
+		loCall = instruction.NewUnconditionalCall(moduleOf[s.lo], loID, s.lo <= MAX_STATIC_RANGE_WIDTH)
```
DavePearce: I see how `NewUnconditionalCall` is being used here, but we can just have a function in this file which switches between `CALL` and `MEMORY_READ` based on the check `s.lo <= MAX_STATIC_RANGE_WIDTH`.

pkg/zkc/vm/internal/transform/range_constraints.go
```diff
@@ -0,0 +1,329 @@
...
+	return fmt.Sprintf("range_u%d", w)
+}
+
+func addRangeCalls[W word.Word[W]](cfg field.Config, modules []Module, rangeModules []Module) []Module {
+	for _, mod := range modules {
+		for _, r := range mod.Registers() {
+			if !r.IsConst() {
+				// TODO : add a call to the corresponding range module for this register.
+				// Issue:
+				// - for register > u16 we need a call that populates the range_uX modules. How to do it ?
```
DavePearce: **Let me think**.  _What is the problem here?_ **First idea**: do all the calls at the beginning of the function.  Ok, this fails for local registers and the return value.  **Second idea**: do it before any `return` instructions.  This works for one-line functions, but not multi-line instructions.  For one-line instructions, we can optimise to avoid multiple calls.  This is done by having a single `ret` at the end of the vector and rewriting any other `ret` to a skip to that one `ret`.  Then, inserting the calls before the `ret`.  For example, this `x = 0; skip_if x==y 2; r=0; ret; r=1; ret` is first rewritten into this: `x = 0; skip_if x==y 2; r=0; skip 1; r=1; ret` and then this: `x = 0; skip_if x==y 2; r=0; skip 1; r=1; $u16[x]; $u16[y]; $u16[r]; ret`.

Hmmmm, it is a bit tricky.  They need to be called once per vector instruction.  So, yes, in principle it can be done by inserting them into each vector instruction, such that they are always run.  This can mean multiple calls though to the same function at different points.  We need a copy of the calls for each distinct "exit path" from the vector instruction.  For example, a vector instruction like `x = 0; skip_if x==y 1; ret; jmp 1` has two exit paths.  It would need to be rewritten as: `x = 0; skip_if x==y 3; $u16[x]; $u16[y]; ret; $u16[x]; $u16[y]; jmp 1`.

Overall, one-line functions are much nicer.  Multi-line functions can get ugly, which is perhaps one more reason to avoid them.  Also, whilst I agree this does seem somewhat complicated --- I still think it makes sense.  For example, tracing with `gogen` will need the function calls and should (optionally) check that e.g. `$u16[x]` does not fail.  I don't see any way around it.

DavePearce: > and when lowering the unconditional call at the --air level it keeps in mind that it has to
> populate the callee module during a sort of trace expansion, like in zkasm ...

Hmmmmm, well I think its at least worth considering this.  Having `gogen` do this for us by executing it as a normal function is by far simplest (despite the technical issues above).  Otherwise, we post-process the trace and then do some kind of separate map-reduce for that (as you say like what is done in zkASM).

The issue is that, for `KOALABEAR`, we never have any type-proof functions to trace.  Thinking, thinking thinking.

Ok, yeah.  **I still like the approach you are taking here.**  The reason being that we want to optimise the lookups.  That means we need explicit checks in the code which can be subsequently optimised away.  We can then run with and without the checks and be sure they are the same.

Its is also worth noting that the bytecode instruction set has an explicit primitive already called `CHECKCAST`.  This is not being used for the same reasons, but it is already being optimised in some cases.  During lowering we would naturally lower `CHECKCAST` into one of your static calls, etc.

DavePearce (changes requested):
Overall, going in the right direction.  Some changes requested.

pkg/zkc/vm/instruction/base.go
letypequividelespoubelles: yes, unconditional stands for call to a non-static uX with x > 16. It's here to flag that we don't need flag :)
good point for static ...

pkg/util/collection/heap/heap.go
```diff
@@ -23,13 +23,21 @@ type Heap[T any] struct {
...
+	var (
+		offset = uint(len(p.data))
+		nsize  = offset + n
+	)
 	// grow data capacity (as necessary)
 	data := slices.Grow(p.data, int(n))
 	// expand data length
 	p.data = data[:nsize]
+	// zero the newly allocated region
+	clear(p.data[offset:])
```
DavePearce, line 41: Hmmmmmm, this is a performance overhead for the bytecode interpreter.  I suggest you split `Alloc` into `AllocRaw` and `AllocAndZero` or `AllocAndClear`.

pkg/zkc/constraints/vector_translator.go
```diff
@@ -87,7 +87,8 @@ func (p *VectorInsnTranslator[F]) translate() Expr[F] {
 		case *instruction.Debug:
 			// no-operation
 			continue
-		case *instruction.Call, *instruction.MemRead, *instruction.MemWrite:
+		case *instruction.Call, *instruction.MemRead, *instruction.MemWrite,
+			*instruction.UnconditionalMemRead, *instruction.UnconditionalCall:
```
DavePearce: Hmmmm, I still don't really like `UnconditionalMemRead` and `UnconditionalCall`.  The name is somehow confusing and doesn't really tell me what's its for [specifically, the name is talking about something which doesn't exist at the WIR level but at the level below].  I would suggest `CastRead` and `CastCall`, but there could be some other names which make sense.  Even `StaticRead` and `StaticCall` would be a little better.  But, meh, not a big drama really.

pkg/zkc/vm/internal/gogen/coverage_test.go
```diff
@@ -64,6 +64,10 @@ var opcodeCoverage = map[string]string{
 	"INT_MULMOD_P": "emitFieldOp (WordTypeF)",
 	// Hint instructions.
 	"HINT_DIVISION": "emitHint (FieldHint)",
+	// Range-check instructions: emitted exactly like their unconditional-free
+	// counterparts (they execute identically; only constraint lowering differs).
+	"UNCONDITIONAL_CALL":        "emitCall (as CALL)",
```
DavePearce: Hmmmm, unsure about that.  Should it not be `emitUnconditionalCall (as UNCONDITIONAL_CALL)`?

pkg/zkc/vm/internal/gogen/emit_memory.go
```diff
@@ -40,6 +40,12 @@ func (g *generator) emitMemRead(c *code, fn *wordFunction, x *instruction.MemRea
 		return fmt.Errorf("gogen: MEMORY_READ from write-once memory %q", mi.name)
 	}
 
+	// A data-less read is a pure (range-check) lookup with no runtime effect,
+	// mirroring executeMemRead's empty-target loop; emit nothing.
+	if len(x.Data()) == 0 {
```
DavePearce, line 45: I think this is OK for now.  But at some point, we'll want the ability to (optionally) emit a bounds check here I guess.

pkg/zkc/vm/internal/machine/base.go
```diff
@@ -280,6 +280,11 @@ func (p *Base[W, I, T]) executeInstruction(insn I, frame StackFrame[W, I],
 	case opcode.CALL:
 		var binsn any = insn
 		return p.executeCall(binsn.(*instruction.Call), frame)
+	case opcode.UNCONDITIONAL_CALL:
+		// Executes exactly like a Call; differs only in constraint lowering.
+		var binsn any = insn
+		c := binsn.(*instruction.UnconditionalCall)
+		return p.executeCall(&instruction.Call{OpIo: c.OpIo}, frame)
```
DavePearce, line 287: You could tidy this up I think by changing `executeCall` to accept the `OpIo` type.  But, not very important.

pkg/zkc/vm/internal/machine/base.go
```diff
@@ -299,6 +304,12 @@ func (p *Base[W, I, T]) executeInstruction(insn I, frame StackFrame[W, I],
 		var binsn any = insn
 		err = p.executeMemRead(binsn.(*instruction.MemRead), frame)
 		// Fall thru
+	case opcode.UNCONDITIONAL_MEMORY_READ:
+		// Executes exactly like a MemRead; differs only in constraint lowering.
+		var binsn any = insn
+		c := binsn.(*instruction.UnconditionalMemRead)
+		err = p.executeMemRead(&instruction.MemRead{OpIo: c.OpIo}, frame)
```
DavePearce: Ditto, and again not very important.

pkg/zkc/vm/internal/transform/range_constraints.go
```diff
@@ -0,0 +1,376 @@
...
+	// are appended after the original modules in ascending width order, so a
+	// width's module id is fixed once the ordering is known.  This lets the
+	// recursive range modules emit calls referencing their (lo/hi) sub-modules.
+	var (
+		base     = uint(len(modules))
+		moduleOf = make(map[uint]uint, len(widths))
+	)
+	//
+	for k, w := range widths {
+		moduleOf[w] = base + uint(k)
```
DavePearce, line 102: Random strategy to use `moduleOf`.  You can just calculate the module id as `base+i` in both loops.

pkg/zkc/vm/internal/transform/range_constraints.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// registerWidth returns the bit-width to range-check a register against, or 0 if
+// it needs no check.  Native (field-element) registers span the whole field
+// [0, modulus) rather than a power-of-two range [0, 2^n), so a bit-range check
+// would be unsound (it would reject valid field elements >= 2^bandwidth); they
+// are therefore not range-checked.
+func registerWidth(cfg field.Config, r register.Register) uint {
+	if r.IsNative() {
+		return 0 //TODO @Dave: do we have to range check the native field registers ?
```
DavePearce: no check for native registers

DavePearce (dismissed):
**Ok, overall I think you've done exactly what I've asked.**  One required change I think is for `Heap.Alloc` as this is imposing (unnecessary) overhead on the bytecode interpreter.  In fact, I don't really know why you've done this?  I guess it is to do with the interpreter, since that's the only place its used.  The need to do this suggests a problem.

**However, I am also thinking about this and what really makes sense**.   There are some problems, and I'm not sure what the right approach is.  Sorry, I just find it hard to think it all through when I'm also deep in other stuff.  Here are my points:

1. (**Unconditional Call**).  So this is actually a problem because its not an unconditional call.  The purpose of having the call is to allow the `range_uXX` functions to be traced like any other.  **But, if the lookup is unconditional, then it may try to lookup values that were not traced**.
2. (**Unconditional Read**).  This differs from the call because it doesn't need to be traced.  For that reason, it is naturally "unconditional".  Hence, I agree it is frustrating to have all the `MemRead` instructions everywhere.

Where does that leave us?  I'm thinking to make `UnconditionalCall` into `CheckCast` or `CallCast`.  Its the same thing, but the lookup will be conditional (we could optimise this more, but its not worth it since we know we're not using it).

Then, for `UnconditionalMemRead`, I'm thinking: _we drop it entirely_.  Instead, when generating constraints, we run a static analysis and identify columns which need type proofs.  Such columns would necessarily be `u16` or below.  The initial analysis could just flag all columns, and we optimise later.  Or you could get claude to write you an analysis which identified those columns needing to be constrained.  Specifically, we could assume parameters were already safe.  The only edge case is for `native` functions where we must constrain the arguments.  Phew! and sorry.

pkg/zkc/vm/internal/transform/range_constraints.go
letypequividelespoubelles: yes

pkg/asm/io/program.go
```diff
@@ -91,7 +91,7 @@ func initialState(registers []Register, buses []Bus, iomap Map) State {
 	)
 	// Initialise arguments
 	for i, reg := range registers {
-		if reg.IsInput() || reg.IsConst() {
+		if reg.IsInput() || reg.IsU1Const() {
```
DavePearce, line 94: Interesting name change.  The intention was to support arbitrary register constants, and I am still expecting to do that to optimise constants at the WIR level.  That's why it was `IsConst()`.  Anyway, not really a big deal and can be changed back if/when we actually support general constants.

pkg/zkc/vm/instruction/base/op_io.go
```diff
@@ -109,7 +109,7 @@ func (p *OpIo) String(mapping SystemMap) string {
 	//
 	switch p.Op {
 	//
-	case opcode.CALL:
+	case opcode.CALL, opcode.UNCONDITIONAL_CALL:
```
DavePearce, line 112: Yup, that looks good.

DavePearce (approved):
LGTM

pkg/asm/io/program.go
letypequividelespoubelles: I do understand the intention (and I thought it was this as well), but it was confusing the name doesn't tell what it's doing.

Issue #1892, opened 2026-06-16
feat: gogen support for mixed arithmetic


DavePearce:
## Summary

Currently, the `gogen` backend fails when enabled in the "mixed tests" for reasons unknown.

ivokub:
@gbotrel - I have currently disabled the gogen-mode tests in monorepo as well, otherwise it blocks merging accelerators which use native arithmetic in zkc. See https://github.com/LFDT-Lineth/lineth-monorepo/pull/3455/changes#diff-ddda5ba12f046a1119848d7077e5991bcd833cf418ac1375a6f984aa4ed1ce78L107-R109. Could you re-enable (or let me know) after it works so that we can re-enable riscv execution tests using gogen mode.

DavePearce:
Fixed as part of #2003

PR #1896, opened 2026-06-16
fix(zkc): generate forces fast mode + gogen support for lowered machine

gbotrel:
Two related gogen fixes.

## 1. `generate` must force fast mode

`zkc generate` regressed after #1882. The fastMode refactor inverted the default: `--fast` defaults off, so `!fastMode` is true and native lowering runs for a command that should never lower.

gogen compiles native ops straight to Go operators. Native lowering produces wide arithmetic (e.g. the 128-bit `0xffff…ffff` mask from `LowerBitwise`) that gogen can't emit:

```
gogen: constant 0xffffffffffffffffffffffffffffffff wider than 64 bits unsupported
```

Fix: force `FastMode(true)` in `applyGenerateDefaults`. gogen consumes the un-lowered, unsplit machine — same reason it already rejects `--split`.

## 2. gogen support for the lowered word machine

So the lowered path works when someone *does* want it (e.g. tracer). Two blockers, found by peeling errors empirically on the RISC-V interpreter (`main.zkc`, KOALABEAR_16):

- **Constants >64 bits** — bitwise lowering emits 128-bit masks. New `constOperand` renders an immediate as an exact ≤128-bit operand (lo/hi pair) so it flows through the arith emitters like any wide value.
- **Wide register as a limb in a multi-register store** — the `shr_u128` lowering helper destructures u128 into `[u1, u127]`. `storePair` is now pair-aware for wide limbs; the narrow distribution paths (`storeNamed`/`storeKnown`) stay explicitly guarded.

Result: the lowered machine generates, `go vet`/`go build` clean, runs. gogen suite green (both shapes).

Not covered: native field-element arithmetic (`gogen: native register unsupported`) — the distinct blocker behind #1892's mixed tests, a separate larger feature.

### DIFF
--- a/pkg/cmd/zkc/generate.go
+++ b/pkg/cmd/zkc/generate.go
@@ -87,6 +87,10 @@ func runGenerateCmd[F field.Element[F]](cmd *cobra.Command, args []string, field
 }
 
 func applyGenerateDefaults[F field.Element[F]](build *BuildConfig[F], quiet bool) {
+	// gogen compiles native ops straight to Go operators, so it must consume the
+	// un-lowered machine: native lowering produces wide arithmetic (e.g. 128-bit
+	// masks) that gogen cannot emit. Force fast mode regardless of the --fast flag.
+	build.config = build.config.FastMode(true)
 	// Suppress printf debug instructions when quiet mode is enabled.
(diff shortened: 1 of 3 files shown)


DavePearce (approved):
LGTM

PR #1905, opened 2026-06-18
feat: support interpreter checkpointing

### LINKED ISSUE #1897: feat: support checkpointing
## Summary

The goal of this feature is to enable the creation of checkpoints with the bytecode interpreter.  The intuition here is that we will have a dedicated "checkpoint" instruction and, when this instruction is executed, it will trigger create a suitable checkpoint and call a user-provided function.  This will require some degree of "coordination" since we won't want to create a checkpoint every time `CHECKPOINT` is encountered but, for example, only every `n` times.

DavePearce:
This adds support for creating checkpoints and restoring them in the bytecode interpreter.  This uses a binary format for representing a checkpoint, where memories are divided into pages (of varying length) to enable a pseudo-sparse representation.  The current implementation supports a notion of a "checkpointing call" instruction.  That is, calls to specific functions are marked as "checkpointing" and, on entry to those calls, can (potentially) generate a checkpoint.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,11 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(go build:*)",
+      "Bash(go test:*)",
+      "Bash(go vet:*)",
+      "Bash(gofmt:*)",
+      "Bash(grep:*)"
+    ]
+  }
+}
--- a/pkg/cmd/zkc/execute.go
+++ b/pkg/cmd/zkc/execute.go
@@ -54,6 +54,9 @@ var executeCmds = []FieldAgnosticCmd{
 	{field.BLS12_377, runExecuteCmd[bls12_377.Element]},
 }
 
+// Permitted flag combinations
+var executeFlags FlagChecks
+
 func runExecuteCmd[F field.Element[F]](cmd *cobra.Command, args []string, field field.Config) {
 	var (
 		errors []error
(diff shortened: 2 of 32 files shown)


.claude/settings.json
letypequividelespoubelles, line 1: not sure it should be pushed :)

pkg/cmd/zkc/execute.go
```diff
@@ -117,7 +146,7 @@ func runExecuteCmd[F field.Element[F]](cmd *cobra.Command, args []string, field
 	// =====================================================
 	// Check Constraints
 	// =====================================================
-	if check && len(errors) == 0 {
+	if check && !resume && len(errors) == 0 {
```
letypequividelespoubelles: check && tracing && len(errors) == 0

instead no ? Easier to understand (for me :) )

pkg/cmd/zkc/execute.go
```diff
@@ -176,6 +205,188 @@ func init() {
...
+	defer closer()
+	//
+	return vm.BootAndExecute(interp, input, 131072)
+}
+
+// parseCheckPointSpec parses a --checkpoint specification, returning the
+// function name, the (positive) interval, and whether the checkpoint is
+// one-shot.  Two forms are accepted:
+//
+//	FUNCTION:INTERVAL  checkpoint on every INTERVAL-th invocation of FUNCTION
```
letypequividelespoubelles, line 246: is it a zkc functiun as an input ? 

I think this not what we want, we'd like a checkpoint with a fct computed by corset like `nbGeneratedCells` (which is not increasing by per word instruction)

pkg/cmd/zkc/execute.go
DavePearce: Yeah, ok fair enough!

pkg/cmd/zkc/execute.go
DavePearce: > I think this not what we want, we'd like a checkpoint with a fct computed by corset

Two things: (1) no, we will be checkpointing based on number of calls to the `interpreter` function; (2) the CLI interface is only for debugging anyway --- this will not determine whatever policy we end up with.

pkg/cmd/zkc/execute.go
DavePearce: In the end, for this I ended up adding in a little framework to tidy up the handling of permitted flag combinations.  Its a bit easier to follow at least.

.claude/settings.json
DavePearce: I did intend specifically to push this.  It just a collection of commands which make sense for this repository which can allow Claude to run autonomously for longer.  Including it means everyone gets to benefit.

letypequividelespoubelles (approved):
LGTM

PR #1914, opened 2026-06-22
fix: assignment to multi-lane memory

### LINKED ISSUE #1913: bug: multi-lane ram assignment
In
```go
func (p *Parser) parseLVal(env Environment) (LVal, []source.SyntaxError) { ... }
```
It seems that `memory` assignments are bugged in that
```go
// construct name symbol
var name = symbol.NewUnresolved(reg, symbol.WRITEABLE_MEMORY, 1)   // the 1 is wrong for multilane RAM
```
expects 1 input parameter rather than the number of expressions in square brackets e.g. 2 for an external access such as `ram[0,0] = 12`.

# Minimal example

```
memory ram(a:u8, b:u8) -> (byte:u8)

fn main<ram>() {
    ram[0,0] = 12
}
```
produces
```
bla:4:5-13 insufficient arguments (expected 2)

    ram[0,0] = 12
    ^^^^^^^^
```

OlivierBBB:


DavePearce (approved):
LGTM

PR #1917, opened 2026-06-22
Minor refactors while reading the code

### LINKED ISSUE #1912: Minor refactors while reading the code

OlivierBBB:


pkg/zkc/compiler/linker.go
```diff
@@ -123,28 +130,26 @@ func (p *Linker) Link() (ast.Program, []source.SyntaxError) {
...
 // identifier and also allocated registers for the address/data lines.
 func (p *Linker) linkDeclaration(index uint) (decl.Resolved, []source.SyntaxError) {
-	switch d := p.components[index].(type) {
+	switch d := p.declarations[index].(type) {
 	case *decl.UnresolvedConstant:
 		return p.linkConstant(*d)
 	case *decl.UnresolvedFunction:
 		return p.linkFunction(*d)
-	case *decl.UnresolvedInclude:
-		return decl.NewInclude[symbol.Resolved](d.Pattern()), nil
```
OlivierBBB, line 147: Unreachable given that `linkDeclaration` is called after filtering out `decl.Include`s

pkg/zkc/compiler/linker.go
```diff
@@ -91,28 +105,21 @@ func (p *Linker) Exists(name string) bool {
...
+// Register a new declaration with this linker. Register saves the declaration's name
+// and adds it to the linker's full slice of Declaration's
+func (p *Linker) Register(declaration decl.Unresolved) {
 	// Ignore any anonymous components (i.e. includes)
-	if component.Name() != "" {
-		// First, record name
-		p.names[component.Name()] = true
-		// Second, act on component type
-		switch c := component.(type) {
-		case decl.Unresolved:
```
OlivierBBB, line 107: This is already the case

pkg/zkc/compiler/linker.go
```diff
@@ -62,25 +65,36 @@ func Link(files ...parser.UnlinkedSourceFile) (ast.Program, source.Maps[any], []
...
+//     UnlinkedSourceFile's (save for include's) into one slice;
+//   - names: records the names of the declarations; used to detect duplicates
+//
+// Together these let the linker resolve every declaration's external
+// identifiers (function, memories, constants, type aliases) into
+// concrete indices, producing the fully linked ast.Program.
+//
+// Note: includes were already followed up on in Compile.
 type Linker struct {
-	busmap     map[string]symbol.Resolved
```
OlivierBBB, line 71: Never used, confirmed by Claude

pkg/zkc/compiler/parser/parser.go
```diff
@@ -2144,14 +2166,24 @@ func (p *Parser) follows(options ...uint) bool {
...
+// [startToken, endToken) from the start of startToken up to (but not
+// including) endToken. This mirrors the half-open convention of source.Span.
+//
+// Note: the final token has index endToken-1
+func (p *Parser) spanOf(startToken, endToken int) source.Span {
 	//
-	start := p.tokens[firstToken].Span.Start()
-	end := p.tokens[lastToken].Span.End()
+	start := p.tokens[startToken].Span.Start()
+	end := p.tokens[endToken-1].Span.End()
```
OlivierBBB, line 2177: Re-using the convention of `[s,e)` as for `Token`, which removes the `p.index-1` that were everywhere.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -1989,19 +1998,26 @@ func (p *Parser) parseLVals(env Environment) ([]LVal, []source.SyntaxError) {
...
 		return lv, errs
-	} else if isDeclared && !p.match(LSQUARE) {
+	}
+	//
+	var (
+		isDeclaredVariable = env.IsDeclaredVariable(reg)
+		lSquareFollows     = p.match(LSQUARE)
+	)
+	//
+	switch {
```
OlivierBBB, line 2016: I found the nested if-else confusing. Also calling `p.match(LSQUARE)` once improves readability (at least for me) since one doesn't have to think of whether `match` will be called once or twice.

pkg/zkc/compiler/ast/decl/memory.go
```diff
@@ -16,32 +16,32 @@ import (
 	"github.com/LFDT-Lineth/zkc/pkg/zkc/compiler/ast/variable"
 )
 
-// MemoryKind determines the type of a given memory (i.e. random access, read
-// only, etc).
+// MemoryKind determines the type of a given memory
+//
+//   - STATIC: static ROM; identical across all executions of a given machine
+//   - ROM: read-only memory; variable across executions
+//   - WOM: write-only and write-once memory; variable across executions
```
letypequividelespoubelles, line 23: I thought it was just write-once memory, but that we can still read it.

pkg/zkc/compiler/parser/environment.go
```diff
@@ -20,6 +20,32 @@ import (
...
+//
+//	for i :u8 = 0; i < 2; i = i + 1 {                  // 1st i
+//		for i :u8 = 4; i < 8; i = i + 2 { ... }    // 2nd i
+//	}
+//
+// due to the first 'i' variable being visible when the 2nd one is being
+// introduced.
+//
+// Note. Creating a block surrounded by braces (within a function body) to
+// to create a new scope is disallowed by the syntax.
```
letypequividelespoubelles, line 48: typo double to

letypequividelespoubelles (dismissed):
LGTM

pkg/zkc/compiler/ast/decl/memory.go
DavePearce: It will be one or the other.  We still have to really decide that.  I've left it open for now.

pkg/zkc/compiler/parser/environment.go
DavePearce: I think there is an open issue to fix this unexpected behaviour.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -283,7 +283,7 @@ func (p *Parser) parseConstant() ([]decl.Unresolved, []source.SyntaxError) {
 	}
 	//
 	component := decl.NewConstant[symbol.Unresolved](name, datatype, constExpr)
-	p.srcmap.Put(component, p.spanOf(start, end-1))
+	p.srcmap.Put(component, p.spanOf(start, end))
```
DavePearce, line 286: In this case, you should still stick with `end - 1` actually.  Maybe add a test case for this example.  For reference, these are the example you should consider:

```
const x:u8 = x
```
```
const y:u8=1, x:u8 = x
```

pkg/zkc/compiler/parser/parser.go
```diff
@@ -306,7 +306,7 @@ func (p *Parser) parseConstant() ([]decl.Unresolved, []source.SyntaxError) {
 		}
 
 		component = decl.NewConstant[symbol.Unresolved](name, datatype, constExpr)
-		p.srcmap.Put(component, p.spanOf(start, end-1))
+		p.srcmap.Put(component, p.spanOf(start, end))
```
DavePearce, line 309: Ditto --- stick with `end - 1`

pkg/zkc/compiler/parser/parser.go
DavePearce: Personally, I don't know that it makes it more readable.  Both score poorly on the readability front IMHO!  But, I don't mind the change.

pkg/zkc/compiler/linker.go
```diff
@@ -36,21 +36,24 @@ import (
 // This can fail for various reasons: for example, if no function of the given
 // name can be found in any source file; or, if a function with the correct name
 // but incorrect arity (i.e. number of parameters/returns) is found.
-func Link(files ...parser.UnlinkedSourceFile) (ast.Program, source.Maps[any], []source.SyntaxError) {
+func Link(unlinkedSourceFiles ...parser.UnlinkedSourceFile) (ast.Program, source.Maps[any], []source.SyntaxError) {
```
DavePearce, line 39: For me, this kind of change doesn't make sense.  `files` is better here because the type already tells you it is an `UnlinkedSourceFile` --- so you're just duplicating information.  But, anyway, I'm fine with the change.

pkg/zkc/compiler/linker.go
DavePearce: Was probably a hang over from zkASM.

pkg/zkc/compiler/linker.go
DavePearce: Nice

DavePearce (dismissed):
Overall, LGTM.  Some nice improvements.  I suggested a further improvement, but not required unless you really want to.

pkg/zkc/compiler/parser/parser.go
OlivierBBB: There is an underlying change to `spanOf` that now has the same half open interval convention `[s,e)` as `Token.Span`. The previous convention for `spanOf` was `[s,e]` (whence all the `end-1` turning into plain `end`). See [this comment](https://github.com/LFDT-Lineth/zkc/pull/1917/changes/BASE..b7727e913af76378c078b72e6d16c11693e35482#r3456089997).

If I use `end-1` we will be using a shorter span than we were previously. Is this what you are suggesting ?

pkg/zkc/compiler/linker.go
OlivierBBB: To each their own. Explicit variable names help me reduce context while reading code, but they make the code more verbose. It would likely be easier for me had I written any of this code myself, but as a reader it helps me a lot.

letypequividelespoubelles (approved):
LGTM

PR #1918, opened 2026-06-23
feat: factorize pc selector

### LINKED ISSUE #1878: guard coming for `PC` flag for multi line function
For multi line function, all the constraints will be guarded with conditions à la `* (- 1 (* (- $pc 1) "(inv (- main:$pc 1))"))`.

Instead, the `--air` level should generate, for a `n>1`multiline functiun, n bit-registers `IS_PC_i`and a constraints:
-  `PC = sum_{i=0}^{n-1} i *IS_PC_i`  
- `~PC = sum_{i=0}^{n-1}  IS_PC_i`

and `IS_PC_i` sould be used to guard each constraints.

letypequividelespoubelles:


pkg/asm/compiler/frame.go
```diff
@@ -31,9 +31,23 @@ func NewAtomicFraming[T any, E Expr[T, E]]() Framing[T, E] {
 	return &OneLineFraming[T, E]{}
 }
 
-// NewMultiLineFraming constructs a suitable framing for a multi-line instruction.
-func NewMultiLineFraming[T any, E Expr[T, E]](pc T, pcWidth uint, ret T, retWidth uint) Framing[T, E] {
-	return &MultiLineFraming[T, E]{pc, pcWidth, ret, retWidth}
+// NewMultiLineFraming constructs a selector-based framing for a multi-line
```
DavePearce, line 34: Good summary!

pkg/asm/compiler/frame.go
```diff
@@ -31,9 +31,23 @@ func NewAtomicFraming[T any, E Expr[T, E]]() Framing[T, E] {
...
+func NewMultiLineFraming[T any, E Expr[T, E]](pc T, pcWidth uint, ret T, retWidth uint,
+	selectors []T) Framing[T, E] {
+	return &MultiLineFraming[T, E]{LegacyMultiLineFraming[T, E]{pc, pcWidth, ret, retWidth}, selectors}
+}
+
+// NewLegacyMultiLineFraming constructs a framing for a multi-line function which
+// guards each instruction using an equality on the program counter.  This is
+// retained for the assembly pipeline, which does not (yet) emit selector
+// registers.
+func NewLegacyMultiLineFraming[T any, E Expr[T, E]](pc T, pcWidth uint, ret T, retWidth uint) Framing[T, E] {
```
DavePearce, line 49: Ok, we keep this so the existing zkasm pipeline continues.

pkg/zkc/constraints/translator.go
```diff
@@ -210,8 +218,28 @@ func initMultiLineFraming[F field.Element[F]](ctx module.Id, pc, ret register.Id
...
+	var (
+		selectorTerms = make([]Expr[F], len(pcSelectors))
+		weightedTerms = make([]Expr[F], len(pcSelectors))
+	)
+	//
+	for c, sel := range pcSelectors {
+		sel_i := mirc.Variable[register.Id, Expr[F]](sel, 1, 0)
+		selectorTerms[c] = sel_i
+		weightedTerms[c] = mirc.Number[register.Id, Expr[F]](uint(c + 1)).Multiply(sel_i)
+	}
```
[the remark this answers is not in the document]
DavePearce: I guess that bugbot is right, but that is because we are not generating lookups yet at all for range constraints ... so its a separate issue.

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -155,12 +158,18 @@ func (p *FullObserver[W, I, M]) assignControlRegisters(m *function.Function[inst
 		nrows = uint(len(states))
 		pc    = uint(len(m.Registers()))
 		ret   = pc + 1
+		// First selector column; one selector per instruction follows.
+		sel = ret + 1
```
DavePearce, line 162: So, the ordering is: `$pc`, `$ret`, $is_pc0` ...

DavePearce (approved):
Basically, looks good to me.  One question I am wondering is whether or not we then even need a `$pc` register?  I guess the one thing that is does is handle padding.  But it could potentially then be a binary column I guess.  Anyway, not necessary to merge this PR.  Maybe something to consider for the future?

pkg/zkc/constraints/translator.go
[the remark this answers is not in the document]
letypequividelespoubelles: yes, it's declared as a u1, so its constraint will be generated later

pkg/asm/compiler/frame.go
letypequividelespoubelles: yes, waiting to rm asm / lisp :)

testdata/zkc/unit/ras.zkc
OlivierBBB, line 1: This zkc file isn't included in any test (e.g. `zkc_unit_test.go`)

pkg/zkc/constraints/translator.go
```diff
@@ -210,8 +218,29 @@ func initMultiLineFraming[F field.Element[F]](ctx module.Id, pc, ret register.Id
...
+	// S = sum of selectors (the activity indicator).
+	sum := mirc.Sum(selectorTerms)
+	// PC == sum_c (c+1)*IS_PC_c (reconstruction)
+	decoding := mir.NewVanishingConstraint("pc_decoding", ctx, util.None[int](),
+		pc_i.Equals(mirc.Sum(weightedTerms)).AsLogical())
+	// PC*S == PC i.e. exactly one selector is high whenever PC!=0 (and, via
```
OlivierBBB, line 238: Here and on line 221 you write that a selector is _high_ rather than _on_ ... On purpose ?

pkg/zkc/constraints/translator.go
```diff
@@ -210,8 +218,29 @@ func initMultiLineFraming[F field.Element[F]](ctx module.Id, pc, ret register.Id
...
+	// pc_decoding, none when PC==0).
+	exclusivity := mir.NewVanishingConstraint("is_pc_exclusivity", ctx, util.None[int](),
+		pc_i.Multiply(sum).Equals(pc_i).AsLogical())
```
OlivierBBB, line 241: Also,
> why `PC * S = PC` rather than `S * S = S` ?

I ask because it's easier for `PC` to vanish than for `S` to vanish. For instance if your base field has modulus `~ 2^(2n-1)` you can have programs of size `2^n` where you can achieve `PC = 0` without `S=0`. This follows from

$$\sum_{0\leq c < 2^n} (c+1) \approx 2^{2n-1}$$

E.g. in Koalabear `2n-1 = 31` has `n = 16`, it would be enough for a function to have `2^16 = 65k` instructions to be able to achieve a weighted sum of the `$is_pc_n` flags, and thus `PC`, that vanishes without `S` (and thus all individual flags) vanishing. And if we ever move to a smaller field, e.g. with a 15 bit prime you could have programs with ~256 instructions that can produce a vanishing weighted sum without forcing all `$is_pc_n` flags to vanish.

With Koalabear it would require roughly a program / function of 2B instructions to have the unweighted sum of `$is_pc_n` flags to vanish, and with a 15 bit prime it would still require 65k instructions in a program to achieve this vanishing of `PC` without `S` vanishing.

Seems like a (minor) soundness issue.

OlivierBBB (commented):
LGTM. I would however change the condition in the constraints from `PC * S = PC` to `S * S = S` for the reason explained in the comments as it seems more future proof.

PR #1922, opened 2026-06-23
feat: lookup constraints for function calls

### LINKED ISSUE #1870: feat: lookup constraints for function calls
## Summary

The purpose of this feature is to generate lookup constraints for function calls.  This is similar to what was done in zkASM, but with some key differences (outlined below).

## Differences

There are two key differences from the proposed approach, compared with how it worked in zkASM:

1. (No `io.Bus`). In zkASM, an `io.Bus` was created for each function called.  This is not necessary in ZkC, since function calls already accept only registers as arguments and returns.  Therefore, a lookup should be created for each distinct function call which operates directly on the argument/return registers in the call instruction.
2. (**OLI Enable Line**) In zkASM, one-line functions did not have an enable line and, instead, supported padding with a valid instance of the function.  In many cases, the tool could not determine a valid instance itself and, hence, it relied on the use to provide hints.  Since this caused problems from a user-interface perspective, the plan for ZkC is to use an enable line `$ret` for OLI functions.

## Approach

The are a few suggested steps:

1. (**OLI Enable**) The first step is to insert a `$ret` line into OLI functions.  Whilst this is not a constant register, it is always `1` in the non-padding region (and `0` in the padding region).  This requires updating the `Framing[F]` to insert the `$ret` line (but no `$pc` line is required), with the `FullObserver` being updated accordingly.
1. (**Always On**) The initial prototype should insert a lookup for each function call, with the `$ret` line being set using a constant register.
1. (**Enable/Disable**). The goal here is to now use a conditional lookup at the call site which is keyed on a binary (`u1`) register.  This register could be a control register, if a suitable one is available.  Otherwise, we introduce a new control register to capture the execution path where the function call is active.

### LINKED ISSUE #1933: bug: y = f(x) + 1 vectorization error
at --wir level, y is written twice in the same vector, need a tmp register to hold lookup values.

letypequividelespoubelles:


pkg/test/zkc_unit_test.go
```diff
@@ -774,21 +774,52 @@ func Test_ZkcUnit_Rem_04(t *testing.T) {
...
+	checkZkcUnit(t, "zkc/unit/call_03", DEFAULT_UNIT_CONFIG.Constraints(true))
 }
 
 // This test covers call register aliasing where the return is assigned to an argument register (x = f(x, y)).
 func Test_ZkcUnit_Call_04(t *testing.T) {
-	checkZkcUnit(t, "zkc/unit/call_04", DEFAULT_UNIT_CONFIG)
+	checkZkcUnit(t, "zkc/unit/call_04", DEFAULT_UNIT_CONFIG.Constraints(true))
+}
+
+// This test performs different module call depending on the branch taken
```
DavePearce, line 794: Better to put comments about what the test is doing inside the test itself as a comment.

pkg/util/logical/proposition.go
```diff
@@ -215,6 +214,15 @@ func simplify[I any, A Atom[I, A]](p Proposition[I, A]) Proposition[I, A] {
 				}
 			}
 		}
+		// Apply consensus: two clauses which agree on every atom except one,
```
DavePearce, line 217: FYI: in the end, I think it will be better if we move away entirely from this algorithm and use a simpler approach rather than trying to and more and more cases to it.  Whilst there are various reasons why I originally used this algorithm in zkASM and, in some ways, it is quite neat: the fact is, it was a bad call and has caused (and continues to cause) actual and difficult-to-resolve problems.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -585,14 +592,24 @@ func (p *StmtCompiler) compileFunctionCall(e *expr.ExternAccess[symbol.Resolved]
 	// Compile arguments
 	arguments, insns := p.compileNonUniformArgs(mapping, e.Args...)
 	if !p.fastMode {
-		// If a register is both an argument and a return of a call (e.g. "x = f(x)"),
-		// snapshot that argument into a fresh temporary first so
-		// the call reads a register distinct from the one it writes.
-		// We could not add this tmp register, but it would imply a lookup with a row shift,
-		// which makes prover's life harder.
-		// We do this only in !fastMode as it is required by the constraints, not by the vm.
+		// The lookup gluing this call to its callee reads the argument and
```
DavePearce: This comment is garbage --- it has absolutely nothing to do with what we are doing here.  Its also confusing because we don't need both the `IsReturn` and `p.assignTargets` cases.  This refactoring is not required for code generation and should be separated out its own transform, with a name like `FlatternCalls()` and appropriate comment explaining why we are actually doing this.

DavePearce: We should never distinguish `fastMode` in the code generator --- it does not need to know anything about this.  It just generates code.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -47,6 +47,9 @@ type StmtCompiler struct {
 	quiet bool
 	// fastMode disables constraint-only rewrites not required by the vm.
 	fastMode bool
+	// assignTargets holds the destination registers of the assignment
+	// currently being compiled.
+	assignTargets map[uint]bool
```
DavePearce: This looks odd to me.  `assignTargets` is only used during a single call to `compileStatement`, and is reset across calls.  Therefore, `StmtCompiler` is not the right place to scope it.

pkg/zkc/constraints/call_lookup.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/LFDT-Lineth/zkc/pkg/util/field"
+	"github.com/LFDT-Lineth/zkc/pkg/util/logical"
+	"github.com/LFDT-Lineth/zkc/pkg/util/word"
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm"
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/instruction"
+)
+
+// moduleInfo captures the metadata about a (potential) callee module which is
+// required to construct a lookup constraint at a call site.
+type moduleInfo struct {
```
DavePearce: Why do this?  Just keep the `*vm.FieldFunction` instances rather than creating yet another struct.

pkg/zkc/constraints/call_lookup.go
```diff
@@ -0,0 +1,371 @@
...
+// The caller (source) side is gated on two (potentially combined) conditions:
+//
+//   - Position: in a multi-line caller the call at code line k fires only on
+//     rows where the selector IS_PC_k is on; in an atomic caller
+//     every row is a call row, so there is no positional gating.
+//   - Path: a call may be executed conditionally. The branch condition under which the
+//     call is reached is materialised by FactorSkipConditions as a boolean
+//     register ("path selector"); only rows where it is on actually call.
+//
+// Lookups require a register (and not an expression) as the source selector,
```
DavePearce, line 48: So, there are two ways to implement the source selector: (1) introduce it using a ZkC transform (meaning it will be filled automatically during tracing); (2) introduce it at the constraints level (meaning it requires a separate tracing mechanism).

This PR adopts option (2) here which makes life difficult with respect to tracing.  Specifically, it creates an assignment in `newPathSelector` which is problematic as assignments will not be used for tracing ZkC in the near future.  All tracing happens through the interpreter (field machine), with the exception of very small number of things (specifically related to the PC).

DavePearce: For option (1) the problem is connecting the introduced selector register with the call itself, such that this can be recovered when generating the lookup.  There are a few ways to do this: (1) use a naming scheme for the selector register which means it can be determined directly from the call instruction; (2) embed the register Id as an `util.Option[register.Id]` in the call instruction itself; (3) something else I didn't think of.

DavePearce: I should add that zkasm also adopted option (1) here for reference.

pkg/zkc/constraints/call_lookup.go
```diff
@@ -0,0 +1,371 @@
...
+	//
+	return selId
+}
+
+// conditionTerm builds a 0/1 term which evaluates to 1 exactly when the given
+// branch condition holds.  Branch conditions are stored in disjunctive normal
+// form over single-bit operands, so this is a pure product/sum of binaries — no
+// normalisation (inverse) is required.  A conjunction is the product of its
+// (0/1) atoms; a disjunction is "1 - product(1 - clause)".
+func conditionTerm[F field.Element[F], T term.Expr[F, T]](cond dfa.BranchCondition, regs []register.Register) T {
```
DavePearce: This section of code is curiously out of place.  Points:

* (**Code Reuse**) We already have code for translating a branch condition into the MIR form, namely: `mirc.TranslateBranchCondition()`.  Therefore, we should be calling that function and avoiding all of this excessive optimisation here (which has nothing to do with generating lookups).  **If there is a feeling that the translation done by that function can be improved, then that function should be improved so all users benefit.**
* (**Code Reuse**) I understand the reason this has been done like this is to parameterize `conditionTerm()` on `T` such that it can be used to generate both an MIR term and a `Computation`.  **However, again, we already have code for converting an MIR expression into a Computation**, and that should be used to avoid all this redundant code.  See `NewComputation()` in `pkg/ir/term/computation.go`.

pkg/zkc/constraints/call_lookup.go
```diff
@@ -0,0 +1,371 @@
...
+// singleBitGuard recognises a branch condition which is, over a single register
+// b, either "b != 0" or "b == 1" — both of which are 1 exactly when the guarded
+// path is taken, so b can serve directly as the lookup selector.  Other forms
+// ("b == 0", "b != 1", a register-valued or non-{0,1} RHS, a multi-register
+// group) cannot, and fall back to a materialised path selector column.
+//
+// Such a call guard is always materialised on a 1-bit register by
+// FactorSkipConditions, so a wider operand here indicates a broken invariant
+// and panics.
+func singleBitGuard(cond dfa.BranchCondition, regs []register.Register) (register.Id, bool) {
```
DavePearce, line 130: I do not particularly like this piece of code, as it is unnecessarily difficult to follow.  It is connected to the issue around tracing the selector see above, and perhaps suggests that using a `util.Option[register.Id]` might be a reasonable approach.  In particular, it would allow this logic to be put into the transform itself.

pkg/zkc/constraints/call_lookup.go
DavePearce: Also, I guess this transform would have to be applied after register splitting as well, but that seems fine.

pkg/zkc/constraints/range_lookup.go
```diff
@@ -0,0 +1,121 @@
...
+// are duplicated.
+const (
+	rangeModulePrefix = "$range_u"
+	rangeValueName    = "value"
+)
+
+// rangeTable locates the static range-check table for a given register width:
+// the id of the module enumerating every value 0 .. 2^n-1, together with the id
+// of its "value" column (the lookup recipient).
+type rangeTable struct {
```
DavePearce, line 41: There is something odd going on with the actual constraint generation which I don't understand.  Specifically, when running this command:

```
> zkc compile --air test.zkc
```

I cannot see the generated static reference tables for the range constraints.  I'm not sure what's going on there, but we definitely want to be to inspect them.

DavePearce: Hmmm, so in `pkg/cmd/zkc/compile.go` I modified `writeIrMemory()` to include this at the end:
```
//
	fmt.Print(")")
	//
	if len(m.Contents()) != 0 {
		fmt.Println(" {")
		fmt.Printf("\t")
		//
		for i, c := range m.Contents() {
			if i != 0 {
				fmt.Printf(", ")
			}
			//
			if i >= 10 {
				fmt.Printf("...")
				break
			}
			//
			fmt.Printf("0x%s", c.Text(16))
		}
		//
		fmt.Println()
		fmt.Println("}")
	}
	//
	fmt.Println()
```
And I can see that with both `--fir` and `--wir` we have the static tables.

pkg/zkc/constraints/range_lookup.go
```diff
@@ -0,0 +1,121 @@
...
+		if !ok || !mem.IsStatic() || !strings.HasPrefix(m.Name(), rangeModulePrefix) {
+			continue
+		}
+		// The "value" column holds the enumerated values and is the lookup target.
+		valId, ok := m.HasRegister(rangeValueName)
+		if !ok {
+			continue
+		}
+		//
+		if w := m.Register(valId).Width(); w <= util.MAX_STATIC_RANGE_WIDTH {
```
DavePearce, line 67: `MAX_STATIC_RANGE_WIDTH` should not be a hard-coded constant.  Rather it should be a parameter to the constraint compiler (i.e. as part of `TraceConfig`).  The default value for it should obviously be what you expect.  This gives us the ability to easily change this should we need to.  Potentially, we should even run tests against a smaller max width (e.g. say `u8`) to make sure the alternate path works. 🤷‍♂️

pkg/zkc/constraints/range_lookup.go
```diff
@@ -0,0 +1,121 @@
...
+	value  register.Id
+}
+
+// computeRangeTables indexes, by register width, the static range-check tables
+// present in the machine.  A static $range_un table fully enumerates the values
+// of an n-bit register and is generated for exactly the widths
+// n <= MAX_STATIC_RANGE_WIDTH; wider registers are range-checked recursively by
+// a call (lowered via addCallLookups), so only the static tables are collected
+// here.
+func computeRangeTables[F field.Element[F]](modules []vm.Module) map[uint]rangeTable {
```
DavePearce: I think the name of this function is misleading.  Perhaps `indexRangeTables` might be better?

pkg/zkc/constraints/translator.go
```diff
@@ -35,10 +35,16 @@ import (
 func GenerateMirConstraints[F field.Element[F]](fm *vm.FieldMachine[F]) mir.Schema[F] {
 	var (
 		modules = make([]mir.Module[F], len(fm.Modules()))
+		// Pre-compute per-module metadata required to wire up call lookups (e.g.
+		// the callee register layout and whether it is atomic).
+		infos = computeModuleInfos(fm.Modules())
```
DavePearce: This is actually ridiculous.  We are recomputing information we already have!!  There is absolutely no need for the `infos` array at all.  Remove it, and just lookup the required information directly in `fm.Modules()` --- that is what is needed.

pkg/zkc/constraints/translator.go
```diff
@@ -147,22 +158,21 @@ func translateFunction[F field.Element[F]](ctx schema.ModuleId, fm vm.FieldFunct
...
 		)
+
 		// Create program counter
-		mod.AddRegisters(register.NewComputed(io.PC_NAME, pcWidth, padding))
+		mod.AddRegisters(register.NewComputed(io.PC_NAME, fm.PcWidth(), padding))
 		// Create return line
 		mod.AddRegisters(register.NewComputed(io.RET_NAME, 1, padding))
-		// Create one-hot program counter selectors
+		// Create IS_PC_<k> program counter selectors
+		pcSelectors = make([]register.Id, len(fm.Code()))
```
DavePearce, line 170: This change is a bit random.  Personally, I thought it was better as it was before!

pkg/zkc/vm/internal/transform/factor_skip_conditions.go
```diff
@@ -78,32 +78,84 @@ func factorSkipConditionsFunction[W word.Word[W]](fn *WordFunction) *WordFunctio
...
 		}
-		//
+		// Nothing to factorize if the condition is a (in)equality on a bit register
+		if isEqualityCondition(si.Cond) && !generatesInverse(si, registers) {
+			factor[uint(i)] = false
+			continue
+		}
+
+		// Factorize if it guards a call, as it is needed for the source selector of the lookup.
+		branchHasCall := branchContainsCall(codes, uint(i), si.Skip)
```
DavePearce: This deeply confused me, and I spent quite a while understanding this.  The comment has no relation to what is really going on.  **Overall, the check against `branchHasCall` should be effectively dead code**.  Specifically, because we should always factor cases for skips which generate inverses.  What is odd though is, in fact, we don't always factor such cases and this is why the `branchContainsCall` is used.  Specifically, we currently have this:

```
if generatesInverse(si, registers) && (elseSize > 0 || thenSize > 1) {
```

So, basically, we currently do not factor a skip when it only contains a single statement.  That goes against the idea of trying to reduce the order of generated constraints, and I don't understand why this exception is made?

DavePearce (changes requested):
Ok, overall, there are a few things to address here.  Firstly, we always want to reuse existing code where possible rather than adding chunks of new (complex) code (Claude seems to have a tendency to do this); Secondly, we must avoid using `schema.Assignment` for tracing at all (e.g. `assignment.NewComputedRegister()`).

For the second point above, its worth noting that the final product will *never* use a `schema.Assignment` for tracing. So any uses of `schema.Assignment` we add now will need to be replaced at some point in the (probably near) future.

pkg/util/logical/proposition.go
DavePearce: I have a proposed approach, FYI, based potentially on some kind of dominators-style algorithm.

pkg/test/zkc_unit_test.go
letypequividelespoubelles: I'd prefer to have it here, you don't have to open all the test to know what they're doing tbf

pkg/zkc/constraints/call_lookup.go
letypequividelespoubelles: because you need to do the computation somewhere, and do it for each module, so better do it once no ?

pkg/zkc/constraints/call_lookup.go
letypequividelespoubelles: >with the exception of very small number of things (specifically related to the PC).

That's exactly the problem and why I did it like this: you need the IS_PC_<k> bit columns, and we already add those columns at the mir level, I don't see the diff with the selector

letypequividelespoubelles: >Also, I guess this transform would have to be applied after register splitting as well,

hum, why ?

pkg/zkc/constraints/call_lookup.go
letypequividelespoubelles: it is made difficylt because skip_if doesn't accept (still) a constant in rhs of the condition. once it's done, it'll become almost trivial as it'll be a product of boolean. there is already a link to the issue to rm the code

pkg/zkc/constraints/range_lookup.go
letypequividelespoubelles: see https://github.com/LFDT-Lineth/zkc/commit/59aae0623820a8cfc2b3bd60caa3f1141075eaf0

you need to add --show-static flag. It's really annoying to have the static tables displayed by default

pkg/zkc/constraints/range_lookup.go
letypequividelespoubelles: I know, already openened an issue for this: https://github.com/LFDT-Lineth/zkc/issues/1937

pkg/zkc/compiler/codegen/statement.go
letypequividelespoubelles: The issue is that it didn't catch:
y = f(x) ; x = x+1.  It was all on the same vector, and as we don't want to have a shift in the lookup, the source of the lookup is re written after the call, thus breaking the lookup. Another option is to change the vectorization rules in !fastMode.

pkg/zkc/compiler/codegen/statement.go
letypequividelespoubelles: >We should never distinguish fastMode in the code generator

It was already the case prior to this PR. We can change it to another transform path if you prefer, but what's the difference ?

pkg/zkc/constraints/translator.go
letypequividelespoubelles: well, I don't remember why I had to do the change, but i can rollback, it's purely cosmetic

pkg/zkc/vm/internal/transform/factor_skip_conditions.go
letypequividelespoubelles: we could do it all the time, do agree. It's just it's not bringing better perf when one single statement; It reduces the degree, but creates an unnecessary register. It's a win only if > 1 statement.

letypequividelespoubelles: done

pkg/zkc/constraints/call_lookup.go
DavePearce: Hmmm, I don't really see that --- the information is already available directly.  You can just cast it.  The issue is its just introducing more code (another `struct`) that mirrors an existing struct we already have.  Its a waste.

pkg/schema/register/register.go
```diff
@@ -224,6 +224,15 @@ func (p Register) Width() uint {
 	return p.width
 }
 
+// WidthOrNative returns the bitwidth of this register, or math.MaxUint if it is a native register.
+func (p Register) WidthOrNative() uint {
```
DavePearce, line 228: Personally, I would go with `util.Option[uint]` here --- but doesn't matter.

pkg/zkc/compiler/codegen/compile.go
```diff
@@ -201,13 +201,22 @@ func (p *Compiler) Compile(declarations []Declaration) (*vm.WordMachine[vm.Uint]
...
-		}
+		wm = vm.FactorSkipConditions[vm.Uint](wm)
+		// FlattenCalls introduces a tmp register to hold a call argument when it's rewritten in the same vector, e.g.:
+		// 1. x = f(x)
+		// 2. y = f(x); x = x + 1
+		// As we want to avoid shift in lookups, we must keep the original value of x in a tmp register,
+		// so that the call can be rewritten as:
+		// 1. tmp = x; x = f(tmp)
+		// 2. tmp = x; y = f(tmp); x = x + 1
+		wm = vm.FlattenCalls[vm.Uint](wm)
```
DavePearce, line 219: Great

pkg/zkc/constraints/call_lookup.go
DavePearce: Ok, I pushed the fix for this --- was pretty easy.

pkg/zkc/constraints/call_lookup.go
DavePearce: > hum, why ?

Yeah, I can't remember my thinking here.  There was some constraint I thought of.  Will see if it comes back.  In the end, it doesn't matter --- if you hit that constraint, then you'll move it down.  And it you don't ... then no worries!!!

pkg/zkc/constraints/range_lookup.go
DavePearce: Perfect

pkg/zkc/constraints/translator.go
DavePearce: I applied this fix, so resolved.

DavePearce (approved):
LGTM.  Note, I did apply the fix to remove use of `moduleInfo`.

PR #1923, opened 2026-06-23
fix: call sharing arg and return register

### LINKED ISSUE #1920: x = call (x) doesn't produce an usable --wir
```
pub input data(address:u8) -> (byte:u8)

fn main() {
    var x:u8 = data[0]
    x = doubleCapped(x)
    if (x != data[1]) {
        fail
    }
}

fn doubleCapped(x:u8) -> (r:u8) {
    r = (x >= 128) ? 255:2 * x
    return
}
```

produces :
```
fn main() {
        u8 x
        u8 $1
        u8 $2
        u8 $3
[0]     $1 = 0x0 ; x = data[$1] ; jmp 1
[1]     x = doubleCapped(x) ; $3 = 0x1 ; $2 = data[$3] ; skipif x == $2 1 ; fail ; ret
}

...
```

but `x = doubleCapped(x)` can't be used as in: `source register == return register` for the constraint generation, as we want to avoid as much as possible to have a shift in the lookup.

letypequividelespoubelles:


pkg/test/zkc_unit_test.go
```diff
@@ -498,6 +498,11 @@ func Test_ZkcUnit_While_03(t *testing.T) {
 	checkZkcUnit(t, "zkc/unit/while_03", DEFAULT_UNIT_CONFIG.Fields(field.BLS12_377))
 }
 
+// x = f(x) in a while loop
```
amkCha, line 501: Thanks for the comment 🙏

amkCha (dismissed):

pkg/test/zkc_unit_test.go
letypequividelespoubelles: yw

DavePearce (approved):
LGTM

PR #1931, opened 2026-06-24
fix: constraint generation with native registers

### LINKED ISSUE #1929: fix: constraint generation for native registers
## Summary

Currently, constraint generation for examples involving native registers leads to a `panic`.

## Example

Minimal test case:

```
pub input data(address:u16) -> (word:u8)

fn main() {
    var lhs:𝔽 = data[0] as 𝔽
    var rhs:𝔽 = data[1] as 𝔽
    var res:𝔽 = data[2] as 𝔽

    if (lhs + rhs) != res {
        fail
    }
}
```

With this command-line:

```
> zkc compile --air basic_01.zkc
```
Gives this outcome:

```
panic: Width() called on native register "lhs"

goroutine 1 [running]:
github.com/LFDT-Lineth/zkc/pkg/schema/register.Register.Width(...)
        /Users/djp/projects/zkevm/go-corset/pkg/schema/register/register.go:221
github.com/LFDT-Lineth/zkc/pkg/zkc/constraints.(*VectorInsnTranslator[...]).ReadRegister(0x104901d80?, {0x14000040501?}, 0x1?)
        /Users/djp/projects/zkevm/go-corset/pkg/zkc/constraints/vector_translator.go:212 +0x24c
```

DavePearce:
This puts in place a straightforward fix for constraint generation in the presence of native registers.  The issues were really around the mishandling of the register width in certain contexts.


pkg/zkc/constraints/vector_translator.go
```diff
@@ -252,3 +253,11 @@ func determineWriteConditions(reg register.Id, branchTable dfa.Result[dfa.Branch
 	//
 	return condition
 }
+
+func bitwidthOf(reg register.Register) uint {
+	if reg.IsNative() {
+		return math.MaxUint
```
letypequividelespoubelles, line 259: ? Can you explain this ? Why not field width ?

DavePearce: Its ugly, but unfortunately the legacy system uses `math.MaxUint` to signal a native register.  This is not a great choice, but also is not that easy to fix right now.

DavePearce: > Why not field width ?

Keep in mind also: there is no field width which can be used.  That's because the modulus does not lie on a 2^n boundary.  Thus, any specific bitwidth we choose is either too restrictive, or too loose.

lorenzogentile404 (approved):

PR #1943, opened 2026-06-30
chore: make the max depth of static tables configurable

### LINKED ISSUE #1937: Make the max length of static tables a build parameter

letypequividelespoubelles:


DavePearce (approved):
Overall, it seems fine so I have approved.  There are few things I don't like though, but I didn't come up with a suggestion solution.  In particular, its odd to have the max static depth as part of the `BinaryFile` --- it somehow feels odd to me. Especially, since we have the `WordMachine` which has been constructed already with a given depth.  Its like we have the same information twice.  Anyway, I'll have to think what the best option is (potentially, just having a more specific set of config options for the `BinaryFile` might help).  But, for now, its fine.

PR #1953, opened 2026-07-02
Disable static reference table tracing

### LINKED ISSUE #1952: Disable tracing of static reference tables
PR https://github.com/LFDT-Lineth/zkc/pull/1834 was merged with static reference table tracing still enabled which should be removed as per https://github.com/LFDT-Lineth/zkc/pull/1834#pullrequestreview-4615647969

OlivierBBB:


OlivierBBB:
Tests fail locally, too.

PR #1959, opened 2026-07-06
feat: add padding

letypequividelespoubelles:


pkg/ir/trace_builder.go
```diff
@@ -266,12 +266,22 @@ func addSpillageAndDefensivePadding[F field.Element[F]](defensive bool, tr *trac
...
 		// Don't pad unless we have to
 		if padding > 0 {
-			// Pad extract rows with 0
+			// Some modules (e.g. one-line functions) have no way to deactivate a
+			// padding row, so instead of a constant padding value they replicate a
+			// real row — which is a valid witness for their constraints.
+			//
+			// TODO: an empty copy-row module has no row to replicate, so it falls
+			// back to the constant padding value, which is generally not a valid
+			// witness.  Such modules need a synthesized/precomputed padding row.
```
DavePearce, line 280: > Such modules need a synthesized/precomputed padding row.

So, the question is where this comes from.

pkg/trace/array_trace.go
```diff
@@ -260,6 +260,23 @@ func (p *ArrayModule[W]) Resize() {
 	p.height = nsize
 }
 
+// UseLastRowAsPadding sets each (already filled) column's padding value to its
```
DavePearce, line 263: Is there a specific reason to choose the final row?  It doesn't matter, I'm just wondering.

pkg/zkc/constraints/call_lookup.go
```diff
@@ -271,15 +271,22 @@ func emitCallLookup[F field.Element[F]](mod *schema.Table[F, mir.Constraint[F]],
 		tgtTerms = registerAccesses[F](callee.Registers(), tgtIds)
 	)
 	//
-	if callee.IsAtomic() {
+	if callee.IsAtomic() && !callee.ContainsMemoryAccess() {
```
DavePearce, line 274: Hmmmmm, interesting one.  We didn't have memory in zkASM, so I didn't come across this case before.

pkg/zkc/vm/internal/trace/full_observer.go
```diff
@@ -166,13 +168,37 @@ func numAuxColumns[W word.Word[W]](m machine.Module) uint {
...
 		return extraColumnsForAccessOnceMemory[W](m)
 	default:
 		return 0
 	}
 }
 
+// assignActivityColumn fills the single $ret activity column of an atomic
+// memory-touching function with 1 on every (active) row and returns its name.
+// Padding rows are left at the column's zero padding value, so $ret==0 there.
+func (p *FullObserver[W, I, M]) assignActivityColumn(m machine.Module,
```
DavePearce, line 184: FYI --- I am rebuilding the FullObserver so this will change somewhat.  Specifically, it will be moved into the `constraints` package as a standalone "post-processing" function.

DavePearce (commented):
Overall, it seems reasonable from my perspective --- though it is getting complicated.  The only concern I have is that I don't see any solution to the empty module case.

pkg/trace/array_trace.go
letypequividelespoubelles: no

pkg/zkc/constraints/call_lookup.go
letypequividelespoubelles: yes, added testcase for OLI modules calling memory

letypequividelespoubelles:
https://github.com/LFDT-Lineth/zkc/pull/1976 is the nuclear option: $ret for all OLI modules

PR #1962, opened 2026-07-06
feat: enforce semantics for random access memory ram

### LINKED ISSUE #1955: feat: enforce semantics for Random Access Memory (RAM)

OlivierBBB:


pkg/zkc/vm/internal/interpreter/interpreter.go
```diff
@@ -310,10 +310,31 @@ func (p *Interpreter[W]) CheckPoint() checkpoint.CheckPoint[W] {
...
 	var (
 		moduleId, _ = p.program.HasModule(mem.Name())
-		page        = checkpoint.NewPage(0, slices.Clone(mem.Contents()))
+		page        checkpoint.Page[W]
+		clock       uint64
 	)
+	// A random-access memory is captured with its per-cell timestamps and clock,
+	// so the timestamps survive the checkpoint.  Other flat memories (ROM/WOM)
+	// have no timestamps and are captured by value only.
+	if ram, ok := mem.(*memory.RandomAccess[W]); ok {
```
DavePearce: This could be pulled out into a separate function like for paged memory, but not a big deal.

pkg/zkc/vm/internal/interpreter/interpreter.go
```diff
@@ -380,18 +402,36 @@ func (p *Interpreter[W]) Restore(cp checkpoint.CheckPoint[W]) {
...
 	if mem.IsReadOnly() {
-		mem.Initialise(flattenMemory(m.Pages()))
+		mem.Initialise(flattenMemory(snapshot.Pages()))
+		return
+	}
+	// Read/write memories (paged or not) restore their timestamped cells and
+	// clock directly -- Reset to a clean slate, then re-install the captured
+	// contents -- so per-cell timestamps survive the checkpoint (rather than
+	// being re-stamped by a replay of writes).
+	switch mem := mem.(type) {
```
DavePearce, line 417: It would have made sense to have the `if mem.IsReadOnly()` part of this switch rather than separate.

pkg/zkc/vm/internal/interpreter/memory_log.go
```diff
@@ -0,0 +1,107 @@
...
+	Accesses() []AccessData[W]
+}
+
+// AccessData records a single memory access: the address, the value and
+// timestamp the cell holds after the access (its write-side), and whether the
+// access was a write.
+type AccessData[W util.Uinter64] struct {
+	address          uint64
+	valueWritten     W
+	timestampWritten uint64
```
DavePearce: Personally, I find `valueWritten` and `timestampWritten` to be confusing, as it somehow implies its only for writes.  I would just have `value` and `timestamp`.

pkg/zkc/vm/internal/memory/memory_log.go
```diff
@@ -0,0 +1,107 @@
...
+// TimestampWritten returns the timestamp the cell holds after this access.
+func (a AccessData[W]) TimestampWritten() uint64 { return a.timestampWritten }
+
+// IsWrite reports whether this access was a write (true) or a read (false).
+func (a AccessData[W]) IsWrite() bool { return a.isWrite }
+
+// CheckpointingMemoryLog is the no-op Log used in fast / checkpoint mode:
+// it records nothing.  Checkpointing captures the timestamped cells directly
+// (see RandomAccess.Cells), so no access log is required.
+type CheckpointingMemoryLog[W util.Uinter64] struct{}
```
DavePearce: I do not understand the name here at all.  I would suggest simply "EmptyMemoryLog" or `NullMemoryLog` or something like that.

pkg/zkc/vm/internal/interpreter/paged_ram.go
```diff
@@ -46,16 +47,21 @@ type PagedRandomAccess[W util.Uinter64] struct {
 	kind     Kind
 	geometry Geometry[W]
 	name     string
-	// Pages of memory, allocated on demand.
-	pages [][]W
+	// Pages of timestamped cells, allocated on demand.
+	pages [][]TimestampedCell[W]
+	// timestamp; increments by 1 with every memory access
+	timestamp uint64
```
DavePearce, line 49: You used the term `clock` above in the checkpoints, and personally I prefer that.  Somehow, its not a timestamp we have here --- rather, its something which makes timestamps.

DavePearce: Not exactly a big deal

pkg/zkc/vm/internal/interpreter/paged_ram.go
```diff
@@ -46,16 +47,21 @@ type PagedRandomAccess[W util.Uinter64] struct {
...
+	// logs memory accesses in chronological order (no-op unless tracing)
+	accessLog Log[W]
```
DavePearce, line 51: So, you didn't fully realise what I suggested.  That is, to use a generic parameter here for performance.  However, I think I would suggest just leaving it as is, and we can address it later if we want.

pkg/zkc/vm/internal/interpreter/paged_ram.go
```diff
@@ -105,54 +111,113 @@ func (p *PagedRandomAccess[W]) Geometry() Geometry[W] {
...
+	//
+	return err
+}
+
+// access performs a single read or write, records it in the access log, and
+// returns the value read from the cell.  Every access re-stamps the touched
+// cell with the current clock; writes always materialise the cell, whilst reads
+// only re-stamp a cell whose page already exists (a read never allocates a page,
+// preserving sparsity).
+func (p *PagedRandomAccess[W]) access(address uint64, isWrite bool, newValue ...W) (W, error) {
```
DavePearce: From a performance perspective, this is a move in the wrong direction.  I think again for now, just leave as is and we can adjust later.

pkg/zkc/vm/internal/interpreter/ram.go
```diff
@@ -13,38 +13,277 @@
...
+	//
+	return err
+}
+
+// access performs a single read or write, records it in the access log, and
+// returns the value that was read from the cell.  Following the memory-
+// consistency argument (see ram.md), every access reads the cell's current
+// (value, timestamp), then re-stamps the cell with the current time: a write
+// stores the new value, a read stores the value back unchanged.
+func (ram *RandomAccess[W]) access(address uint64, isWrite bool, newValue ...W) (W, error) {
```
DavePearce: Again, it should be clear that this negatively impacts performance.

pkg/zkc/vm/internal/memory/static_array.go
```diff
@@ -31,76 +31,76 @@ import (
 // The type parameter W is the word type (e.g. a field element or big.Int), and
 // D is the AddressDecoder strategy that encodes the layout of rows within the
 // flat slice.
-type StaticArray[W util.Uinter64] struct {
+type StaticArray[W util.Uinter64, C any] struct {
 	kind     Kind
 	geometry Geometry[W]
```
DavePearce: This will conflict now :(

ram.md
```diff
@@ -0,0 +1,209 @@
+# RAM module
```
DavePearce: Not sure what this is doing here --- looks like more of a working document than something we should be checking in.

DavePearce (changes requested):
Overall, this basically ready to merge.  The only required change is to remove the `ram.md` file which looks like some kind of working notes.  There are some other changes I've suggested to improve the code, but these are minor --- up to you.  And, finally, I have made a few comments around performance --- but my advice is to leave those and we'll fix them separately.

pkg/zkc/vm/internal/interpreter/paged_ram.go
OlivierBBB: OK

ram.md
DavePearce: This file does not belong in the root directory.  Personally, it looks more like some working comments / thoughts than something we want to check in.  However, you could maybe put it in the `docs/` directory at least.

pkg/zkc/vm/internal/interpreter/ram.go
```diff
@@ -13,39 +13,279 @@
...
+	// Raising the timestamp initially means every deliberate access has a
+	// nonzero timestamp, so touched cells are identifiable in the finalization
+	// phase.
+	ram.timestamp++
+	//
+	value := ram.valueAt(address)
+	// A read stores the value back unchanged, re-stamped with this access's time.
+	ram.restamp(address, value)
+	//
+	if ram.logging {
```
DavePearce, line 52: This definitely looks much better, and more like what I was expecting.

Unfortunately, there is still a problem with this code because it does not increment the timestamp correctly.  Specifically, it increments the timestamp for every cell in a memory --- not every row.  So, for memories with multiple data lines --- it gets the wrong timestamp. 

I suggest the easiest way to do this is to have a method `Clock()` which increments `timestamp`, and you then call that from within the `Interpreter`.  Then, inside `Read` and `Write` you don't increment `timestamp`.  Anyway, this is just a suggestion ---- there are other solutions.

split-bug-investigation.md
```diff
@@ -0,0 +1,388 @@
+# Register-splitting bug: subtraction & division on KOALABEAR_16
```
DavePearce: Same as for `ram.md` --- does not belong in the root directory.

DavePearce (changes requested):
Ok, looking much improved.  Still requesting changes though, as there a bug in the `timestamp` calculation.  Also you still have the `ram.md` and another `md` file in the root directory.

ram.md
OlivierBBB: Hm, I did remove this file ... I don't understand why it's still here.

DavePearce (approved):
Ok, LGTM  --- this is ready to merge.

PR #1967, opened 2026-07-07
feat: update range proof of u1 registers

### LINKED ISSUE #1910: No static table for range_u1 proof
add `r * ( 1 - r) ==. 0` instead

### LINKED ISSUE #1970: fix: bug in handling of range check functions
## Summary

Generate range check functions are not implemented correctly, as they insert manual memory reads --- which they do not need to do.  This actually breaks the (somewhat quirky) use of static reference tables for range checks.  There is also an additional issue where an error is reported that the range check module cannot be found.

## Example 1

Minimal example:

```
pub input data(address:u16) -> (byte:u16)

fn main() {
    var x:u16 = data[0]
}
```

Then, compiling with this command, we can see the problem:
```
> zkc compile --max_static_depth=256 test.zkc
```

Which produces:

```
pub input data(address:u16) -> (byte:u16)

fn main() -> ()
         ...

static $range_u8() -> (value:u8)
   {0x0, 0x1, 0x2, 0x3, 0x4, 0x5, ...}

fn $range_u16(value:u16) -> ()
         u8 lo
         u8 hi
 [00]  mov hi::lo = value
       read  = $range_u8[lo]
       read  = $range_u8[hi]
       ret
```

Here, we can see the problematic reads (e.g. `read = $range_u8[lo]`).  Such reads are not required as lookups are (currently) added automatically.  For example, running this command on our original program:

```
> zkc compile test.zkc
```

And we get this:

```
pub input data(address:u16) -> (byte:u16)

fn main() -> ()
         u16 x
         u16 $1
 [00]  ldc $1 = 0x0
       read x = data[$1]
       ret

static $range_u16(index:u16) -> (value:u16)
   {0x0, 0x1, 0x2, 0x3, 0x4, 0x5, ...}
```

Noting that this is inconsistent since it does not insert reads to `$range_u8[lo]`

## Example 2

Now, running the following command on our example from above:

```
> zkc compile --max_static_depth=256 --mir test.zkc
```

And we get the following error:

```
panic: missing static range table for width 16
```

letypequividelespoubelles:


pkg/test/zkc_unit_test.go
```diff
@@ -1148,6 +1150,47 @@ func Test_ZkcUnit_Padding_05(t *testing.T) {
 	checkZkcUnit(t, "zkc/unit/padding_05", util.DEFAULT_CONFIG)
 }
 
+// ===================================================================
+// Range check Tests
+// ===================================================================
+// Range check a u16, exercised across several static-table depth limits so
+// that the width is covered by a static table (depth 2^16), and by recursive
+// destructuring (depths 2^8 and 2^4).
+func Test_ZkcUnit_RangeCheck_01(t *testing.T) {
```
DavePearce, line 1157: So, I think this is OK ... but again my feeling would be to run every test we have with different values of `MaxStaticDepth`.  Basically, I just don't know why we wouldn't do that.

testdata/zkc/unit/range_check_01.zkc
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
DavePearce, line 3: Ok, I don't like this --- non-executing tests are pretty pointless.  Maybe you forgot to add the `accepts` or `rejects` files?

pkg/test/zkc_unit_test.go
```diff
@@ -1148,6 +1150,47 @@ func Test_ZkcUnit_Padding_05(t *testing.T) {
...
+	checkZkcUnitStaticDepths(t, "zkc/unit/range_check_01",
+		DEFAULT_UNIT_CONFIG, codegen.DEFAULT_MAX_STATIC_DEPTH, 1<<8, 1<<4)
```
DavePearce: Since the depths are the same I would just put them into `checkZkcUnitStaticDepths`.

DavePearce (changes requested):
Ok, overall, I think this looks good.  The new `RangeCheck_01` tests don't have any input data which seems odd to me?  Also, personally, I would just run all the tests with different static depths (or, at least, all the unit tests).

testdata/zkc/unit/range_check_01.zkc
letypequividelespoubelles: Well, I wanted the minimal to test the air compilation. If we add more things it leads to a casting tests, and involve more registers.

pkg/test/zkc_unit_test.go
letypequividelespoubelles: We could ... But it will*3 the CI time

DavePearce (approved):
Yup, LGTM

PR #1972, opened 2026-07-08
feat: support fast-mode splitting

### LINKED ISSUE #1968: feat: support splitting of shift and div/rem
# Summary

The div/rem operations and shift left/right do not currently support register splitting.  Whilst these operations are only present in fast mode, it is still nice to be able to support fast-mode splitting to enable support for large (e.g. `u256`) arithmetic, such as required for an EVM.

DavePearce:
This applies a wide range of fixes related to register splitting, and adds support for fast mode splitting (which, for example, allow big number arithmetic in fastmode).


OlivierBBB (approved):
Blind approaval

PR #1976, opened 2026-07-09
feat: add pdding for one line functiun (nuclear option)

### LINKED ISSUE #1761: feat: padding for one-line functions in ZkC constraints
## Summary

Currently, there is an issue related to one-line functions in constraints generated from ZkC functions.  Specifically, they do not support padding as they do not have any way to initialise the padding values.  In zkASM this was done using a "witness evaluation" step, but that proved problematic for lots of reasons (specifically, having to manually specify the witness).  There are two plausible approaches.

## Approaches

* (**Dedicated Return**) The obvious solution would be to use a `$ret` along the lines of that done for multi-line functions, with lookups always being conditional on this.  Every constraint would then be guarded by this.
* (**Trace Extraction**) This approach is basically the same as what was done originally in zkASM, but with a twist. Padding values are not statically determined but, instead, extracted from the trace itself.  So, there are two scenarios regarding the trace of a module `m`: **(a)** empty trace; or **(b)** non-empty trace.  For the non-empty trace, its easy --- we just exact the first line of the trace to use as the padding values (this is similar to how shifted inversions was handled for legacy constraints).  For the empty trace, we cannot extract anything obviously.  **Therefore, we simply require that the prover does not pad empty modules**.

Provided agreement can be reached on the padding of empty modules, then the second approach here seems preferable.

letypequividelespoubelles:


pkg/zkc/constraints/post/process_function.go
```diff
@@ -29,23 +29,53 @@ type Transcriber[W Word[W], F Element[F]] func(state vm.State[W]) []F
...
+// Here, REGS is the set of registers declared by the given function.  The RET
+// activity line is 1 on every non-padding row, since each instance of a one-line
+// function occupies exactly one active row; padding rows are left at 0.
+// TODO: see https://github.com/LFDT-Lineth/zkc/issues/1975
+// OLI won't have a $ret column
 func ProcessOneLineFunction[W Word[W], F Element[F]](f vm.Function[W], states []vm.State[W]) rtrace.ArrayModule[F] {
 	var (
-		regs = array.Map(f.Registers(), toRtraceRegister)
-		rows = transcribe(states, oneLineTranscriber[W, F])
+		hasRet = !f.IsNative()
```
DavePearce, line 46: Interesting choice to not have a return line.  I guess this is something we will have talk over with Ivo.

pkg/zkc/constraints/post/process_function.go
```diff
@@ -29,23 +29,53 @@ type Transcriber[W Word[W], F Element[F]] func(state vm.State[W]) []F
...
+			copyState(st, row)
+
+			return row
+		}
+		// Reserve an extra column for the $ret activity line.
+		var row = make([]F, width+1)
+		// Copy over function state
+		copyState(st, row)
+		// Every real row of a one-line function is active.
+		row[width] = one
```
DavePearce, line 75: Looks neat!

pkg/zkc/constraints/post/process_function.go
```diff
@@ -56,9 +86,9 @@ func oneLineTranscriber[W Word[W], F Element[F]](st vm.State[W]) []F {
 // (which requires adding control lines as necessary).  The format of a trace
 // row for a multi-line function is:
 //
-// +----------+----+-----+--------+--------+-----+
-// |   REGS   | PC | RET | IS_PC0 | IS_PC1 | ... |
-// +----------+----+-----+--------+--------+-----+
+// +----------+-----+----+--------+--------+-----+
```
DavePearce, line 89: Ok, you swapped the order of the `RET` / `PC` line so its consistent for both multi-line and one-line functions.  Makes sense 👍

DavePearce (approved):
Overall, LGTM.  No real comments, so seems good to go!

pkg/zkc/constraints/post/process_function.go
letypequividelespoubelles: Yes, I don't exactly know how the interface will be. It'll be super easy to add the $ret for native if we have to

PR #1979, opened 2026-07-09
feat: switch constraint generator over to bytecode api

### LINKED ISSUE #1978: feat: switch constraint generator over to bytecode API
# Summary

The purpose of this issue is to refactor the constraint generator to operate on the bytecode API, thereby avoiding the need to convert into a field machine, etc.  This should be relatively straightforward given how similar the bytecode API is to the legacy API.

DavePearce:
Translate MIR/AIR constraints directly from the bytecode program
(descriptor modules + bytecode vectors) instead of routing through the
legacy word/field machine (BytecodeProgramToWord + WordToFieldMachine).

The word→field lowering of arithmetic/concat instructions is folded into
the translator, and the same dfa branch-table/write-map analysis now
runs over bytecode vectors. Concrete bytecode instruction structs are
re-exported from pkg/zkc/vm so the translator can dispatch on them
across the internal package boundary.


letypequividelespoubelles (approved):
LGTM

PR #1981, opened 2026-07-10
test: add padding to zkc test

### LINKED ISSUE #1980: add padding in zkc tests
# Summary

The testing framework for legacy constraints included dedicated support for testing different levels of padding.  Specifically, it would run each test separately with `0 ... n` rows of padding on each module.  The purpose of this is, of course, to test that generated constraints still work correctly under padding.  **The purpose of this issue is to implement a similar system for ZkC**.

letypequividelespoubelles:


pkg/cmd/corset/check.go
```diff
@@ -150,8 +148,9 @@ type CheckConfig struct {
 	// Specifies whether to use Coverage testing and, if so, where to write the
 	// Coverage data.
 	Coverage util.Option[string]
-	// Specifies the range of Padding values to check
-	Padding util.Pair[uint, uint]
+	// Specifies whether to expand each module's length up to the next power of
```
DavePearce: Ok, you're making a very specific choice here.  It is not at all clear to me that this is a good choice, and that is why I did it differently before.  However, perhaps what we could do is to run the test two ways: once with a small number of padding levels (e.g. 0..5) and then also apply the "pad to power" test.

pkg/ir/trace_builder.go
```diff
@@ -49,12 +50,12 @@ type TraceBuilder[F field.Element[F]] struct {
...
 	// amount.
 	checks bool
-	// Determines the amount of padding to apply to each module in the trace.
-	// At the moment, this is applied uniformly across all modules.  This is
-	// somewhat cumbersome, and it would make sense to support different
-	// protocols.  For example, one obvious protocol is to expand a module's
-	// length upto a power-of-two.
-	padding uint
+	// Determines whether or not padding is applied to each module in the trace.
+	// When enabled, every module's length is expanded (with front padding
```
DavePearce: I suggest a sensible rewording is:

> When enabled, various padding checks are applied (e.g. expanding every module's length upto a power of two).

pkg/ir/trace_builder.go
```diff
@@ -116,8 +117,9 @@ func (tb TraceBuilder[F]) WithValidation(flag bool) TraceBuilder[F] {
 	return ntb
 }
 
-// WithPadding updates a given builder configuration to use a given amount of padding
-func (tb TraceBuilder[F]) WithPadding(padding uint) TraceBuilder[F] {
+// WithPadding updates a given builder configuration to control padding.  When
+// enabled, every module's length is expanded up to the next power of two.
+func (tb TraceBuilder[F]) WithPadding(padding bool) TraceBuilder[F] {
```
DavePearce: Yeah, this is definitely not what we want.  We must be able to specify precise padding values, but also ok support padding to power of 2.  Thinking about it, we need an interfacecalled something like: `PaddingPolicy` with a function like `DeterminePadding(module.Id, uint) uint` which, given a trace length and a module identifier, returns the amount of front padding to apply.  Then, we change to `WithPadding(PaddingPolicy)` and that then determines the padding each module gets.  This allows us to set different policies quite easily.

pkg/test/util/check_valid.go
```diff
@@ -64,6 +65,9 @@ type Config struct {
 	// enable quiet mode, which elides printf statements and calls to #[debug]
 	// functions during code generation.
 	quiet bool
+	// enable padding of the generated trace, expanding every module's length up
+	// to the next power of two.
+	addPadding bool
```
DavePearce: This also does not make complete sense to me.  Every test should be run for padding values e.g. `0,1,2` and also using the power-of-2 policy --- no exceptions.  I guess, if there are tests which currently fail when padding is enabled, then having a flag so those tests can disable padding (until we fix them in some later PR) make sense.  In which case, padding should be enabled by default.

pkg/test/util/check_valid.go
```diff
@@ -386,16 +398,16 @@ func bootAndCheckpoint[W vm.Word[W]](t *testing.T, program vm.Program[W], tc Tes
 	return program, checkpoints, outputs
 }
 
-func runConstraintTest(t *testing.T, p vm.Program[vm.Uint], test TestCase, cfg codegen.Config) {
+func runConstraintTest(t *testing.T, p vm.Program[vm.Uint], test TestCase, cfg codegen.Config, addPadding bool) {
```
DavePearce: Here, instead of accepting the `addPadding` boolean you could accept a `PaddingPolicy` object.

DavePearce (changes requested):
Ok, I understand what you are trying to do here, but I am not happy with the execution.  Minimum things we need: (1) testing a range of (small) fixed padding values (e.g. `0`,`1`,`2`) as well as for the "power-of-two" policy; (2) ability to specify fixed padding value on the command-line.  I have therefore suggested using a `PaddingPolicy` interface to provide the necessary flexibility to support these (and potentially other) padding modes.

pkg/ir/padding_strategy.go
```diff
@@ -0,0 +1,46 @@
+// Copyright Consensys Software Inc.
```
DavePearce, line 1: I guess this is the best place this for now, since it covers both legacy and zkc.

pkg/test/util/check_legacy.go
```diff
@@ -250,11 +252,16 @@ func checkTraces[F field.Element[F]](t *testing.T, test string, maxPadding uint,
 
 func checkTrace[F field.Element[F], C sc.Constraint[F]](t *testing.T, tf lt.TraceFile, id traceId,
 	schema sc.Schema[F, C], mapping module.LimbsMap) {
+	// Map the legacy padding toggle onto a padding strategy.
+	paddingStrategy := ir.NoPadding
+	if id.padding {
```
DavePearce, line 257: Its not exactly an improvement as you're not testing the existing strategy, just: nothing or the next power of two strategy.  I guess for legacy we don't care so much.

pkg/test/zkc_unit_test.go
```diff
@@ -1182,17 +1182,43 @@ func Test_ZkcUnit_RangeCheck_05(t *testing.T) {
...
-var STATIC_DEPTHS = []uint{codegen.DEFAULT_MAX_STATIC_DEPTH, 256}
+var STATIC_DEPTHS = []uint{256, 1 << 12}
+
+// ZKC_PADDING_STRATEGIES enumerates the padding strategies that every ZkC unit
+// test is exercised against (see checkZkcUnit).
+var ZKC_PADDING_STRATEGIES = []struct {
+	name     string
+	strategy ir.PaddingStrategy
+}{
+	{"no-padding", ir.NoPadding},
```
DavePearce: Well, its definitely an improvement.  But basically, in most cases, it will mean testing just `0` or `1` additional line of padding.

pkg/ir/trace_builder.go
```diff
@@ -326,14 +322,43 @@ func checkModuleHeights[F field.Element[F]](original []uint, defensive bool, tr
...
-		tr.Pad(i, padding*multiplier, 0)
+		var (
+			height     = tr.Module(i).Height()
+			multiplier = tr.Module(i).Name().Multiplier
+			logical    = height / multiplier
+			target     uint
+		)
+		//
+		switch strategy {
+		case SingleRowPadding:
```
DavePearce: Ok, so this is working ... but the implementation could be much better.  Basically, to add a new strategy, we have to update the `padding_strategy.go` file and also this code.  It would be much nicer to just have to update `padding_strategy.go`.  You can do this quite easily by changing `PaddingStrategy` to have a type like this: `type PaddingStrategy func(height, multiplier uint) uint`.  Then `PADDING_STRATEGIES` is still a `map[string]PaddingStrategy` but, instead of putting constants into it, you put functions.

DavePearce (dismissed):
Ok, basically it looks fine.  But actually the way you handle the strategies could be improved a lot using either lambda functions or interfaces.  I guess lambda functions is the easiest.

pkg/ir/trace_builder.go
DavePearce: Also, it would be good to know, for the given test suite we have, how often `target = util_math.NextPowerOfTwo(logical) * multiplier` returns a `target` where `target != height`.  Because I guess it will be quite often, and maybe testing `1` and `2` would make sense over `0` and `1` (unsure).

DavePearce (dismissed):
LGTM

DavePearce (approved):
LGTM

PR #1986, opened 2026-07-13
fix: Claude fix for subtraction

### LINKED ISSUE #1985: bug: reusing a register across two split-assignments (::) in a #[native] function produces an out-of-range register reference in the generated lookup constraint
### Summary

When a `#[native]` function contains two separate split-assignments (`hi::lo = ...`) that write into the *same* high-limb ("overflow") register, the compiled binary's AIR schema contains a lookup constraint whose selector references a register index equal to `numRegisters` for that module (i.e. one past the last valid index). Any code that walks/prints the constraint (e.g. calling `.Lisp()`) panics with an out-of-range index. This was found via `zkc compile --split`, but the bad register reference is baked into the emitted binary constraints, not just the debug-printing path, so it looks like a register-allocation bug rather than a purely cosmetic one.

Using two distinct registers for the two split-assignments (e.g. `overflow1`, `overflow2` instead of reusing `overflow`) avoids the issue entirely, with the same call structure otherwise.

### Environment

- `zkc`/go-corset: local checkout used via `replace` in a downstream Go module (`github.com/LFDT-Lineth/zkc v1.2.20 => ../../go-corset`)
- Compile command: `zkc compile --output out.bin --split repro.zkc`
- Field: default (`KOALABEAR_16`)

### Minimal repro

`repro.zkc`:

```
type tLimb              = u16
type tLimbAddr          = u8
type tWidth             = u64
type tWidthLarge        = u128
const nbLimbs:tLimbAddr = 4

memory lhs(address:tLimbAddr) -> (limb:tLimb)
memory rhs(address:tLimbAddr) -> (limb:tLimb)
memory mod(address:tLimbAddr) -> (limb:tLimb)
memory res(address:tLimbAddr) -> (limb:tLimb)

pub input in_lhs(address:tLimbAddr) -> (limb:tLimb)
pub input in_rhs(address:tLimbAddr) -> (limb:tLimb)
pub input in_mod(address:tLimbAddr) -> (limb:tLimb)
pub input in_res(address:tLimbAddr) -> (limb:tLimb)

fn main<lhs,rhs, mod, res>() {
    var i:tLimbAddr
    for i = 0; i < nbLimbs; i = i + 1 {
        lhs[i] = in_lhs[i]
        rhs[i] = in_rhs[i]
        mod[i] = in_mod[i]
    }
    mulmod_64()
}

#[native]
fn mulmod_64<lhs,rhs,mod,res>() {
    var lhsL:tWidth = lhs[0]::lhs[1]::lhs[2]::lhs[3]
    var rhsL:tWidth = rhs[0]::rhs[1]::rhs[2]::rhs[3]
    var modL:tWidth = mod[0]::mod[1]::mod[2]::mod[3]

    var product:tWidthLarge = (lhsL as tWidthLarge) * (rhsL as tWidthLarge)

    var overflow:tWidth
    var quotient:tWidth
    overflow::quotient = product / (modL as tWidthLarge)

    var remainder:tWidth
    // Reusing `overflow` here (instead of a fresh register) triggers the bug.
    overflow::remainder = product - ((quotient as tWidthLarge) * (modL as tWidthLarge))

    var res0:tLimb
    var res1:tLimb
    var res2:tLimb
    var res3:tLimb
    res0::res1::res2::res3 = remainder
    res[0] = res0
    res[1] = res1
    res[2] = res2
    res[3] = res3
}
```

Compile:

```sh
zkc compile --output repro.bin --split repro.zkc
```

Compilation itself succeeds without error or warning.

Then, a small Go program that just deserializes the binary and prints each AIR constraint (no downstream tooling involved):

```go
package main

import (
	"fmt"
	"os"

	"github.com/LFDT-Lineth/zkc/pkg/util/field/koalabear"
	"github.com/LFDT-Lineth/zkc/pkg/zkc/constraints"
)

func main() {
	data, err := os.ReadFile(os.Args[1])
	if err != nil {
		panic(err)
	}
	var binf constraints.BinaryFile[koalabear.Element]
	if err := binf.UnmarshalBinary(data); err != nil {
		panic(err)
	}
	sc := binf.AirConstraints()
	css := sc.Constraints().Collect()
	for i, cs := range css {
		_ = cs.Lisp(&sc).String(false)
		_ = i
	}
	fmt.Println("done")
}
```

```sh
go run . repro.bin
```

### Actual result

```
panic: runtime error: index out of range [876] with length 875

goroutine 1 [running]:
github.com/LFDT-Lineth/zkc/pkg/schema.(*Table[...]).Register(...)
	pkg/schema/module.go:227
github.com/LFDT-Lineth/zkc/pkg/ir/term.(*RegisterAccess[...]).Lisp(...)
	pkg/ir/term/register_access.go:168
github.com/LFDT-Lineth/zkc/pkg/schema/constraint/lookup.(*Vector[...]).Lisp(...)
	pkg/schema/constraint/lookup/vector.go:113
github.com/LFDT-Lineth/zkc/pkg/schema/constraint/lookup.Constraint[...].Lisp(...)
	pkg/schema/constraint/lookup/constraint.go:176
github.com/LFDT-Lineth/zkc/pkg/ir/air.Air[...].Lisp(...)
	pkg/ir/air/constraint.go:169
```

The panicking constraint is a lookup constraint whose target module is `mulmod_64`, and whose selector is a `RegisterAccess` pointing at register index `numRegisters` for that module (one past the last valid register index — off by exactly 1). `mulmod_64`'s module reports `875` registers (valid indices `0..874`), but the selector references index `876`.

### Expected result

Either:
- Compilation should reject reusing a register across two split-assignments within the same native function (clear compile-time error), or
- Register allocation/lookup-constraint generation should not produce an out-of-range register reference — the generated constraints should be internally consistent with the module's register count regardless of register reuse.

### Notes / isolation

- This reproduces purely by deserializing the compiled binary and calling `.Lisp()` on the AIR constraints — no downstream driver code is involved.
- The number of call sites of `mulmod_64()` (one vs. two) does not affect reproduction; the trigger is specifically reusing one register (`overflow`) as the discarded high limb of two separate split-assignments within the native function body.
- Compiling the same program but with two distinct registers (e.g. `overflow1` for the quotient split and `overflow2` for the remainder split) compiles and prints cleanly with no panic, with everything else identical.
- Because the bad register index is present in the emitted binary constraints file (not just recomputed each time `--air` is requested from source), this looks like a genuine register-allocation/lookup-generation bug in the compiler rather than a printing-only issue, and may affect soundness of the resulting constraint system, not just diagnostics.

OlivierBBB:
Claude generated PR summary

# Summary

zkc exec --split on a small field (e.g. KOALABEAR_16) produced bit overflow errors for chained subtractions and for
division. Two independent bugs in the bytecode interpreter path were responsible; both are fixed.

```bash
# before: both errored with "bit overflow"
zkc exec --field KOALABEAR_16 --split break.json   break.zkc     # division
zkc exec --field KOALABEAR_16 --split break_2.json break_2.zkc   # x - y - z
```

```bash
#after
break.zkc   → out = 0x00001234
break_2.zkc → out = 0x00007ffe
```

##  Bug 1 — split subtraction borrow

Split subtraction reused addition's carry mechanism, storing the two's-complement sign-extension bits of a (possibly
negative) chunk result as the "borrow". That only equals the true borrow for a single-bit borrow (one subtrahend).
A chained subtraction (a - b - c) can borrow more than one 2^regWidth unit, so the next limb went out of range.
Division is affected because it lowers to x - q·y - r (a chained subtraction into a u0 assert register).

Fix: new pre-split pass DecomposeSubtractions rewrites any to-be-split subtraction with ≥2 negative terms into a
chain of single-negative-term subtractions, each of which splits with a correct 1-bit borrow. It only touches
subtractions wide enough to actually be split, so small wrapping subtractions keep their two's-complement semantics.

##  Bug 2 — multi-limb DIV_HINT operand order

DIV_HINT read/wrote multi-limb operands least-significant-limb-first, but the splitter lays limbs out big-endian
(most-significant limb at the lowest register index). So dividend/quotient/remainder came out limb-swapped once an
operand spanned more than one limb.

Fix: loadHintOperand / storeHintResult now iterate big-endian, matching the splitter's layout. Single-limb (unsplit)
behaviour is unchanged.

# Changes

- pkg/zkc/vm/internal/transform/decompose_subtraction.go — new pass
- pkg/zkc/vm/transform.go, pkg/zkc/compiler/codegen/compile.go — wrapper + wiring (before SplitRegisters)
- pkg/zkc/vm/internal/interpreter/interpreter.go — hint limb order
- pkg/test/zkc_unit_test.go + testdata/zkc/unit/{sub_01,div_06}.* — regression tests

# Tests

make zkc-test passes (0 failures). New tests run the bytecode interpreter only, because the constraint and gogen
paths do not yet support split subtract-with-borrow / multi-limb hints — a pre-existing limitation already noted for
basic_45 ("TODO: subtract with borrow"). Extending those paths is left as follow-up.


pkg/zkc/vm/internal/transform/decompose_subtraction.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/word"
+)
+
+// DecomposeSubtractions rewrites every integer subtraction having more than one
+// "negative term" (i.e. more than one subtrahend register, or a subtrahend plus
+// a non-zero constant) into a chain of subtractions, each having exactly one
+// negative term.  For example (all u32):
+//
+//	z = x - y - w        =>   t = x - y ; z = t - w
+//	z = x - y - w - 1    =>   t0 = x - y ; t1 = t0 - w ; z = t1 - 1
```
OlivierBBB, line 31: ### Finding 6 — evaluated "sum-then-subtract" alternative to decomposition (REJECTED on measurement)

Question raised: the current fix decomposes `a - b - c - d` into a left-to-right chain
`t0 = a-b ; t1 = t0-c ; z = t1-d`, creating n−1 intermediate registers (each a trace
column). Proposed alternative: sum all subtrahends (+const) with ONE addition, then a single
subtraction — `s = b+c+d ; z = a-s`. Rationale: addition splits with just a carry, not a
chain, and `a - s` is single-subtrahend (1-bit borrow, already handled).

Measured register count of `main` (KOALABEAR_16, --split, after SplitRegisters, before range
constraints) for `out = in0 - in1 - ... ` with N subtrahends, via a temporary ZKC_DUMP hook:

```
| N subtrahends | decomposition (shipped) | sum-then-subtract |
|---------------|-------------------------|-------------------|
| 2             | **16**                  | 20                |
| 3             | **22**                  | 23                |
| 4             | 28                      | **26**            |
| 5             | 34                      | **29**            |
| 6             | 40                      | **32**            |
| 7             | 46                      | **35**            |
| 8             | 52                      | **38**            |
| 16            | 100                     | **62**            |
```

Reference points: a pure ADD of 5 addends adds only **1** computed register (one carry);
a single `a - b` (u32) is 10 registers total.

Interpretation:
- Decomposition grows ~linearly (~6N): each step adds a narrow (operand-width) 2-limb temp
  plus a borrow.
- Sum-then-subtract has a lower slope but higher constant: the sum `s = Σsub` is *wider* than
  the operands (e.g. b+c of two u32 needs 33 bits → 3 limbs) and needs its own carries, so for
  small N it costs MORE than decomposition's narrow temps. Crossover is ~N=3–4.
- **Common case is N=2** (division lowers to `z0 = x - q*y - r` and `z1 = y - r - w - 1`;
  typical user subtractions are 2 terms). There decomposition is *smaller* (16 vs 20).

**Decision: keep the decomposition.** Per the "only switch if it saves registers" criterion,
sum-then-subtract does not save for the realistic workload (it regresses N=2). It would only
help for many-subtrahend subtractions, which are rare. Reverted the experimental
sum-then-subtract implementation; shipped code remains the left-to-right decomposition.

If a future workload is dominated by high-arity subtractions, revisit: a *hybrid* (sum-then-
subtract only when N is large) would capture the tail without regressing N=2, at the cost of
extra complexity.

DavePearce:
Completed as part of other PRs.

PR #1993, opened 2026-07-16
feat: support bytecode validation

### LINKED ISSUE #1991: feat: support bytecode validation checks
# Summary

The purpose of bytecode validation is to provide checks which ensure every bytecode is valid within the given environment after the various transforms have been applied.  Specifically, to check any bytecode specific invariants (e.g. that source registers are valid, widths all make sense, etc).

DavePearce:
This adds support for validating the bytecode program at the end of code generation.  This is to catch any malformed bytecodes before the program is passed into subsequent phases.


letypequividelespoubelles (approved):
LGTM

PR #1998, opened 2026-07-17
fix: subtract with borrow

### LINKED ISSUE #1766: Create new "borrow substraction" vm instruction
So, at the moment, its pretty unpleasant in terms of the number of registers being created.  This is really coming from limitations in the `vm` instruction set.  For example, we should have an instruction "subtract with borrow" which allows us to do the subtraction directly rather than shifting it right by 1.  Also, better support for constants, no need for cast instructions, etc.

_Originally posted by @DavePearce in https://github.com/Consensys/go-corset/pull/1750#discussion_r3231521477_

DavePearce:
This puts through various fixes for the "subtract with borrow" semantics, and also refactors `split.Addition` and `split.Concat`.


pkg/test/zkc_bench_test.go
```diff
@@ -37,15 +37,16 @@ func Test_ZkcBench_FastPow(t *testing.T) {
 }
 
 func Test_ZkcBench_Gcd(t *testing.T) {
-	checkZkcBench(t, "zkc/bench/gcd", DEFAULT_BENCH_CONFIG.GoGen(false))
+	checkZkcBench(t, "zkc/bench/gcd", DEFAULT_BENCH_CONFIG)
 }
 
 func Test_ZkcBench_Fnv1aHash(t *testing.T) {
-	checkZkcBench(t, "zkc/bench/fnv1a_hash", DEFAULT_BENCH_CONFIG.Constraints(false))
+	// TODO: unknown error
```
letypequividelespoubelles: there is an error ? But the test is still running wo CI failure ? I don't understand the TODO

pkg/test/zkc_bench_test.go
```diff
@@ -82,33 +83,34 @@ func Test_ZkcBench_Poseidon_U32_Small(t *testing.T) {
...
 		Splitting(false).Constraints(false).GoGen(false))
 }
-func Test_ZkcBench_Poseidon_U32_Big(t *testing.T) {
-	// TODO: investigate test failures
-	checkZkcBench(t, "zkc/bench/poseidon/test/u32/poseidon_zeros_big_01", DEFAULT_BENCH_CONFIG.
-		Splitting(false).Constraints(false).GoGen(false))
-}
+
+// func Test_ZkcBench_Poseidon_U32_Big(t *testing.T) {
+// 	// TODO: investigate test failures
```
letypequividelespoubelles: please open issues to not forget the TODOs ...

letypequividelespoubelles (commented):
LGTM

pkg/test/zkc_bench_test.go
DavePearce: It was a problem which was fixed, and I should have removed the `TODO`.  Done now.

pkg/test/zkc_bench_test.go
DavePearce: Ok, I have updated this to refer to #2005 which is an active issue / PR.

gbotrel (approved):

PR #2001, opened 2026-07-17
feat: optimize and or xor

### LINKED ISSUE #1747: ref tables for AND / XOR / OR
Currently the zkc constraints generates a cascade of module AND_64 -> AND_32 -> AND_16 -> AND_8 -> AND_4 -> AND_2 -> AND_1 which creates a lot of tiny modules. We should be able to (dynamically, at compile time) stop before reaching U1)

### LINKED ISSUE #1877: inefficiencies in bitwise operation lowering and register splitting
Let's assume we're in `KOALABEAR_16`, and we perform a  `a:u32 & b:u32` operation in `main`.
It will create :
- `a:u32` is plited in two registers
- `main` is calling `and_u32`
- `and_u32` is splitting `in_1` and `in_2` in two `u16` registers to call two times `and_u16`

It would make more sense to call directly `and_u16` on each sub registers of main. Impact:
- no `and_u32` modules (saves 6 splitted registers + lookups and inner constraints)
- 1 more flag on `main`(two lookups instead of one)
- no change for `and_u16`

It seems a clear win.

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/split/bitwise.go
```diff
@@ -123,7 +123,10 @@ func newOrBytecode[W word.Word[W]](bitwidth uint, target RegisterId, sources ...
 
 func newXorBytecode[W word.Word[W]](bitwidth uint, target RegisterId, sources ...RegisterId) Bytecode[W] {
 	if len(sources) != 2 {
-		return bytecode.NewBitwise[W](bytecode.OP_NOT, target, sources[0], sources[0], util.Cast[uint16](bitwidth))
+		// Surplus limb of the wider operand: the narrower operand contributes a
```
DavePearce, line 126: Interesting --- good find

pkg/cmd/zkc/root.go
```diff
@@ -110,7 +110,7 @@ func GetBuildConfig[F field.Element[F]](cmd *cobra.Command, field field.Config)
 	build.config = codegen.DEFAULT_CONFIG.
 		Inlining(GetFlag(cmd, "inline")).
 		FastMode(fastMode).
-		MaxStaticDepth(GetUint(cmd, "max_static_depth")).
+		MaxStaticDepth(GetUint(cmd, "max-static-depth")).
```
DavePearce, line 113: Thank you

DavePearce (approved):
Yeah, Looks Good.  I got quite confused for a moment, as I was wondering why we didn't have any lookups being generated.  But, of course, that is because there are no lookups generated for read/write!

PR #2003, opened 2026-07-17
feat(zkc): dedicated FieldCast instruction for 𝔽↔uint conversions

gbotrel:
Register splitting passed `FieldArith` through verbatim, so a split wide-uint operand feeding a local field op kept stale register ids:

```
z = (x:u32 as 𝔽) + y   →   fadd y = x'1⊕x'0   // wrong target; x's limbs added as field elements
```

Introduce a dedicated `FieldCast` bytecode as the sole 𝔽↔uint bridge. Every other bytecode is now purely uint or (field arithmetic) purely native, so splitting `FieldArith` is just an id remap:

```
cast $5 = x'0::x'1
fadd z = $5⊕y
```

Semantics are **checked** (not reduce), consistent with narrowing uint casts (which fail rather than truncate):

- `uint → 𝔽` asserts the value is canonical (`< P`) — free when the source width guarantees it.
- `𝔽 → uint` asserts it fits the target width.

This upholds the invariant that a native register always holds a value in `[0, P)`, and is enforced on all three execution paths: the interpreter and gogen check at execution (gogen elides the check when the interval analysis proves it statically), and for the constraint pipeline `LowerFieldCasts` materialises it as generated `$field_range_*` helper functions (limb-wise comparison against the modulus) invoked through the ordinary call mechanism.

gogen also gains native-register support for moduli ≤ 64 bits (a field element is a single uint64 local; the mod-P emitters reduce on use), so the mixed suite — previously interpreter-only — now runs on gogen too.

Tests: `basic_11` (split wide-uint into a local field op), `felt_casting_02` (u32→𝔽→u32 round-trip), `felt_casting_03` (𝔽 parameter/return across a call), each with accepts + rejects; `basic_08`'s over-P accept vector moves to rejects.


pkg/zkc/compiler/codegen/compile.go
```diff
@@ -180,6 +180,8 @@ func (p *Compiler) Compile(declarations []Declaration) (vm.Program[vm.Uint], []s
 		if p.config.splitting {
 			program = vm.SplitRegisters(p.config.field, program)
 		}
+
+		program = vm.LowerFieldCasts(program)
```
DavePearce: This is not what I am expecting --- but lets see.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -478,14 +478,23 @@ func (p *StmtCompiler) compileCast(e *expr.Cast[symbol.Resolved], bitwidth util.
...
-	if bitwidth.IsEmpty() {
+	switch {
+	case bitwidth.IsEmpty() && e_bitwidth.IsEmpty():
+		// 𝔽 → 𝔽 (no-op cast): compile the source directly into the native target.
 		return p.compileExpr(e.Expr, bitwidth, mapping, targets)
-	} else if e_bitwidth.HasValue() && e_bitwidth.Unwrap() <= bitwidth.Unwrap() {
-		// upcast
+	case bitwidth.IsEmpty():
+		source, insns := p.compileUniformArgs(e_bitwidth, mapping, e.Expr)
+		return append(insns, vm.FieldCast[vm.Uint](targets, source))
```
DavePearce: This feels odd to me --- again not what I am expecting.

pkg/zkc/vm/internal/interpreter/interpreter.go
```diff
@@ -917,23 +879,33 @@ func (p *Interpreter[W]) executeFieldMul(pc uint32, codes []uint32, stack []W) (
...
+
+func (p *Interpreter[W]) executeFieldCast(pc uint32, codes []uint32, stack []W) (uint32, error) {
+	var (
+		targets, sources, n = encoding.DecodeRegisterLists(pc, codes)
+		module              = p.program.Module(p.fid)
+		val                 = loadAcross(module, sources, stack)
+	)
+
+	first := targets
+	if module.Register(first.Next()).IsNative() {
```
DavePearce: This is a static check --- it should not be done at runtime.  We should distinguish the two cast forms in the encoding.

pkg/zkc/vm/internal/transform/lower_field_cast.go
```diff
@@ -0,0 +1,183 @@
...
+	"strings"
+
+	"github.com/LFDT-Lineth/zkc/pkg/schema/register"
+	"github.com/LFDT-Lineth/zkc/pkg/util"
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/bytecode"
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/descriptor"
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/word"
+)
+
+// LowerFieldCasts inserts the canonicality checks required by field conversions.
```
DavePearce: this is confusing --- why are these not already added during code generation when we lower from the AST to the VM instructions?

pkg/zkc/vm/internal/transform/lower_field_cast.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil
+	}
+
+	return []Bytecode[W]{bytecode.CallFun[W](uint16(p.ensure(widths)), regs, nil)}
+}
+
+func (p *fieldCastHelpers[W]) ensure(widths []uint) uint {
+	var builder strings.Builder
+
+	builder.WriteString("$field_range")
```
DavePearce, line 120: So, these "field_range" checks are interesting, and not entirely expected.  I guess the question is: perhaps we want a different semantic here.

DavePearce (commented):
So, overall, I think this looks mostly good.  I didn't find any issues with the way the field casts were generated.  However, having finally understood the purpose of the `$field_range` checks ... I have to wonder if there is not an easier way.  For example, we could change the semantic of the cast to prevent the need for a range change?

DavePearce (dismissed):
LGTM

DavePearce (approved):
LGTM

PR #2006, opened 2026-07-20
feat: tidy up poseidon tests

### LINKED ISSUE #2005: fix: refactor poseidon test
# Summary

Currently, the poseidon test has two versions `u32` and `felt` --- but we only need to support the `felt` version.  Also the poseidon tests do not make use of test vectors, like other benchmark tests.

DavePearce:
This tidies up the way the poseidon tests operate.  Specifically, they now use dynamic test vectors rather than hard-coded individual tests. Furthermore, the u32 version of poseidon is dropped since this does not reflect the intended use case of poseidon.


OlivierBBB (approved):
LGTM

PR #2012, opened 2026-07-20
feat: delete all ref to debug function in quiet mode

### LINKED ISSUE #2011: `--quiet` should completely kill [#debug] module, not only kill the calls

letypequividelespoubelles:


DavePearce (approved):
Ok, so I am happy to merge this as it is working and not causing any issues.  But, for the record, there is a mechanism (not yet implemented) by which this is supposed to happen.  Specifically, the function `Prune()` in `descriptor.Program` is intended (once implemented) to remove all functions which cannot be reached from the `main` entry point.  Thus, this would also include `#[debug]` functions when `quiet` was enabled as they would not be reachable.

PR #2017, opened 2026-07-21
chore: default to quiet mode

letypequividelespoubelles:


DavePearce (approved):
LGTM

pkg/cmd/zkc/root.go
```diff
@@ -136,7 +136,7 @@ func init() {
 	//
 	rootCmd.PersistentFlags().Bool("show-static", false, "Show static tables in the MIR/AIR output")
 	rootCmd.PersistentFlags().BoolP("fast", "f", false, "Fast-mode execution (no tracing, no constraints)")
-	rootCmd.PersistentFlags().BoolP("quiet", "q", false, "suppress debug output")
+	rootCmd.PersistentFlags().BoolP("quiet", "q", true, "suppress debug output")
```
DavePearce: I actually realise after looking at this again, that I don't like it.  It means to run the program and see output, you need to run e.g. `zkc exec --quiet=false`.  What we need is a modifier which allows us to see output.  I might suggest "debug".  So, for example, `zkc exec -d` would turn off "quiet mode".

DavePearce (changes requested):

DavePearce (dismissed):
LGTM

amkCha (approved):

PR #2019, opened 2026-07-21
feat: add lookup for mem read / write

### LINKED ISSUE #1936: reuse already existing lookup selector
When lowering wir to mir, when we generate a fresh 1-bit selector for the lookup, we should first check if we can use an existing one (will be the case if several calls in the same branch table / row)

### LINKED ISSUE #2013: feat: generate lookups for SROM/ROM/WOM
# Summary

Currently, there are no lookups being generated for memory reads to/from SROM, ROM or WOMs.  The purpose of this feature is to implement this.  The key challenge is around handling of lookup selectors, since both ROM and WOMs have selectors to prevent lookups into the padding region.

letypequividelespoubelles:


pkg/zkc/constraints/call_and_memory_lookup.go
```diff
@@ -31,68 +31,127 @@ import (
...
-// callSourceSelector determines the register used to gate the source (caller)
-// side of a (conditional) call's lookup constraint, given the branch condition
-// under which the call executes and the caller's per-line is_pc_* selectors
-// (empty for an atomic caller).
-func callSourceSelector[F field.Element[F]](mod *schema.Table[F, mir.Constraint[F]], ctx schema.ModuleId,
+// lookupSourceSelector determines the register used to gate the source
+// (accessing) side of a (conditional) call's or memory access's lookup
+// constraint, given the branch condition under which the access executes and
+// the accessor's per-line is_pc_* selectors (empty for an atomic function).
+func lookupSourceSelector[F field.Element[F]](mod *schema.Table[F, mir.Constraint[F]], ctx schema.ModuleId,
```
DavePearce, line 153: Hmmm, is there an issue with this function?  Should it not return `register.Id` rather than `util.Option[register.Id]`?.  In other words, it must always find a source selector.

pkg/zkc/constraints/call_and_memory_lookup.go
```diff
@@ -293,6 +352,80 @@ func emitCallLookup[W vm.Word[W], F field.Element[F]](mod *schema.Table[F, mir.C
...
+		tgtIds = make([]register.Id, len(srcIds))
+	)
+	//
+	for i := range tgtIds {
+		tgtIds[i] = register.NewId(uint(i))
+	}
+	// Build the source (accessor) vector.
+	var source mir.LookupVector[F]
+
+	if srcSelector.HasValue() {
```
DavePearce: So, this.  How can we have `!srcSelector.HasValue()`?  I don't see how this is possible.  I think you can either just use `Unwrap()` directly or (as suggested above) change `srcSelector` so its a `register.Id`, but not an `util.Option[register.Id]`.

DavePearce: The reason I'm picking on this is because I'm trying to understand how we get to the `lookup.UnfilteredVector()` case ... which seems broken to me.

pkg/zkc/constraints/call_and_memory_lookup.go
DavePearce: Specifically, the issue is that this program:

```
pub input data(address:u16) -> (byte:u32)
pub output res(address:u16) -> (byte:u32)

fn main() {
    var lhs:u32 = data[0]
    var rhs:u32 = data[1]
    if lhs != rhs {
        res[0] = f(lhs,rhs)
    }
}

fn f(lhs:u32, rhs:u32) -> (r:u32) {
   r = lhs + rhs
}
```

Generates this lookup (which AFAIK) is wrong:

```
(lookup "call_2_0_3"
   ((_ f:lhs'1 f:lhs'0 f:rhs'1 f:rhs'0 f:r'1 f:r'0))
   ((main:$lookup_sel_12 main:lhs'1 main:lhs'0 main:rhs'1 main:rhs'0 main:$5'1 main:$5'0)))
```

Specifically, it has no target selector ... so it can lookup into the padding region.

DavePearce (changes requested):
Overall, I think it looks good and I was going to merge.  But, I think I found an issue with the call lookup, where I managed to generate one which doesn't have a target selector.  Given our current position on handling of one-line functions ... I guess that seems broken ?

pkg/zkc/constraints/call_and_memory_lookup.go
letypequividelespoubelles: issue creating while a rebase of https://github.com/LFDT-Lineth/zkc/pull/1979/

fixed in this PR.

DavePearce (approved):
Yup, LGTM.

PR #2028, opened 2026-07-22
fix: bypass mul splitting when not required

### LINKED ISSUE #2015: fix: unnecessary multiplication splitting
# Summary

There appear to be some issues related to the multiplication splitting algorithm, of which this issue identifies one.

# Example

The following illustrates a minimal problematic case:

```
fn mul(x:u8, y:u8) -> (z:u16) {
   z = (x * y) as u16
}
```

Which currently generates the following IR:

```
fn mul(x:u8,y:u8) -> (z:u16)
         ...
 [00]  mul pp$4::pp$3 = x*y
       mov m$5 = pp$3
       mov m$6 = pp$4
       mov m$7 = m$5
       mov m$8 = m$6
       cat z = m$8::m$7
       ret
```

This clearly doesn't make sense since the result of `x*y` fits into `z` without requiring any splitting.

DavePearce:
Previously, the multiplication splitting algorithm was splitting multiplication even when it was not required to do so (i.e. because the multiplication fitted into the available bandwidth, etc).  This led to some inefficiency for small multiplications.


pkg/zkc/vm/internal/transform/split/multiplication.go
```diff
@@ -54,21 +56,29 @@ func Multiplication[W word.Word[W]](mapping descriptor.LimbsMap[W], alloc Alloca
...
 		g            = mulGranularity(mapping.RegisterWidth(), mapping.BandWidth(), targetWidths...)
+		operands     [][]RegisterId
 	)
 	//
 	one = one.SetUint64(1)
-	// Degenerate case: a product with no source registers is simply its
-	// constant.  Codegen does not normally emit this, but handle it anyway.
+	//
 	if len(insn.Source) == 0 {
+		// Degenerate case: a product with no source registers is simply its
```
letypequividelespoubelles, line 69: shouldn't we panic here instead ?

letypequividelespoubelles (approved):

pkg/zkc/vm/internal/transform/split/multiplication.go
DavePearce: Well, this should be caught as part of the validation stage.

PR #2034, opened 2026-07-23
fix: discrepency around handling of subtract

### LINKED ISSUE #2021: feat: investigate `gogen` subtraction width disconnect
# Summary

There is an odd-looking check added into `gogen` to handle vector subtraction, which seems to suggest potentially a bug in the interpreter (i.e. because this is following the interpreter, but we should not have this check).

## Details

The problematic check is in `emit_arith.go`:

```
// The general vector encoding widens arithmetic to the target width.
// The compact SUBC and SUB_2n1 encodings retain their source-derived
// wrap widths, so only mirror the target widening when SUB_nm is used.
if usesVectorSubEncoding(x) {
	wrap = max(wrap, store.total)
}
```

This suggests a difference in the way that `SUB_2n1` and `SUB_nm` work with respect to the target `bitwidth` --- and that should not be the case.

DavePearce:
There was a discrepency between the way that subtraction instructions were being handled with respect to underflow.  Specifically, vectored subtracts were being handled differently from the non-vectored variations.  This discrepency was mirrored in gogen as well.  The non-vectored subtraction instructions were being handled incorrectly, and that is now fixed --- along with specific tests for this.

In addition, have introduce a notion of "test sampling" which allows one to specify a sampling ratio for a given test.  This is useful when we have a large set of test vectors, but don't want to test all of them on every run.


gbotrel:
Not a big fan of the sampling, it can silently hide regressions and can be subtle to maintain I think .

Is the win really noticeable? Can we reproduce locally using testing.Short flag to run less stuff locally and run more stuff on the CI ?

gbotrel (dismissed):

gbotrel (dismissed):

letypequividelespoubelles (approved):
blind approval

PR #2039, opened 2026-07-23
fix: register splitting of casted skip_if

### LINKED ISSUE #2031: fix: mismatched limbs for `SkipIf`
# Summary

A `panic` arises in certain cases with respect to the `SkipIf` instruction.  This should be straightforward to fix.

## Example

The following test illustrates:

```
pub input data(address:u16) -> (byte:u32)

fn main() {
    var word:u32 = data[0]
    var hi:u16
    var lo:u16
    // destructure
    hi::lo = word
    //
    if (hi as u32) != data[1] || (lo as u32) != data[2] {
        fail
    }
}
```

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/split_registers.go
```diff
@@ -340,14 +340,73 @@ func splitWrite[W word.Word[W]](limbsMap descriptor.LimbsMap[W], alloc split.All
...
+	//
+	var (
+		nlimbs = make([]RegisterId, n)
+		pre    = make([]Bytecode[W], n)
+	)
+	// Allocate a consecutive block of fresh registers, zeroing the padding
+	// limbs and copying the original limbs across.
+	for i := range nlimbs {
+		if i < n-m {
+			nlimbs[i] = alloc.Allocate("z", util.Some[uint](0))
```
DavePearce, line 400: You need to use `alloc.ZeroRegister` here.  This helps to ensure only one zero register is every allocated, and that it is allocated in the correct form so we know its a zero registers.

DavePearce (changes requested):
Overall, I don't think this is the right fix.  We can support `SkipIf` with variable sized operands, and the branch condition translator (see pkg/asm/compiler/branches.go) already handles this case correctly.  In particular, since branch conditions compile down into polynomial constraints, its not efficient to allocate registers to handle differences in limb size.

Fix should be roughly like this:

* Remove check in `bytecode.NewSkipIfVec()`
* Fix encoding problem in encoding for `SkipIf`
* Update `executeSkipIf_rv` to handle differing numbers of limbs
* Possibly update `gogen` accordingly.

PR #2041, opened 2026-07-23
perf: lowering switch

### LINKED ISSUE #1894: optimize wir from switch statement
similar of https://github.com/LFDT-Lineth/zkc/issues/1793 but for `SKIP_MULT` arising from `switch` statement

Each branch should create it's own boolean guard, and the default branch should be guarded by sum (branch_boolean) == 0

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/split/allocator.go
```diff
@@ -40,6 +40,10 @@ type Allocator[W any] interface {
 	// This is presumed to be a computed register, and automatically assigned a
 	// unique name.  No assignment is included for the allocated register
 	Allocate(prefix string, width util.Option[uint]) RegisterId
+	// AllocateNamed allocates a fresh (computed) register of the given width
+	// bearing exactly the given name.  The caller is responsible for ensuring
+	// the name is unique within the target module.
+	AllocateNamed(name string, width util.Option[uint]) RegisterId
```
DavePearce, line 46: No, please don't add more to the interface when we don't need it.  Just use existing `Allocate` function with the name as the prefix.

pkg/zkc/vm/internal/transform/split/switch.go
```diff
@@ -30,3 +30,41 @@ func Switch[W word.Word[W]](mapping descriptor.LimbsMap[W], insn *bytecode.Switc
...
+	//
+	for i, c := range insn.Cases {
+		ncases[i] = bytecode.DispatchCase{Bit: onlyLimb(mapping, c.Bit), Skip: c.Skip}
+	}
+	//
+	return []Bytecode[W]{&bytecode.Dispatch[W]{Cases: ncases, Default: onlyLimb(mapping, insn.Default)}}
+}
+
+// CheckCast splits a checkcast instruction.
+func CheckCast[W word.Word[W]](limbsMap descriptor.LimbsMap[W], c *bytecode.CheckCast[W],
```
DavePearce: This is in the wrong place.

DavePearce: But, what does it mean?  That `CheckCast` instructions are now being added before splitting?  Why would that be?

pkg/zkc/vm/internal/transform/lower_switch.go
```diff
@@ -103,42 +118,134 @@ func lowerSwitchVector[W word.Word[W]](vec BytecodeVector[W], registers split.Al
...
+	codes = append(codes,
+		// one = 1
+		bytecode.LoadConst(onereg, one),
+		// sum = b_0 + ... + b_{n-1}
+		bytecode.AddConst(sumreg, bits, zero),
+		// b_default = one - sum
+		bytecode.NewArith(bytecode.OP_SUB, []descriptor.RegisterId{bdef},
+			[]descriptor.RegisterId{onereg, sumreg}, zero),
+		// Explicit range proof for b_default:
+		bytecode.NewCheckCast[W](bdef, 1))
```
DavePearce: This looks wrong to me.  The checkcast operation should be added as part of the `InsertCheckCasts` transform.  This is exactly the kind of situation that is already handled in `insertCheckCasts`, so I don't see any reason to deviate from that.

pkg/zkc/vm/internal/transform/split/allocator.go
DavePearce: One option is to change the algorithm for determining the name from the prefix in order to reduce lots of indicates being added.  That would be reasonable, and an improvement.  Otherwise, we end up in a situation where some columns have this naming scheme, and others have another naming scheme.  Its not ideal, and it could hide serious soundness issues.

pkg/zkc/vm/internal/transform/split/switch.go
DavePearce: So, I see that `CheckCast` is being added during `LowerSwitch`, but this goes against the convention used by other transforms.

pkg/zkc/vm/internal/transform/lower_switch.go
DavePearce: My analysis (see the issue notes) is pretty clear: you do not need this checkcast operation because `bdef` is a `u1`.  A cast should be inserted automatically by `InsertCheckCasts`, and if that is not the case then there is a bug in the transform [which can be fixed separately].

pkg/zkc/vm/internal/transform/lower_switch.go
```diff
@@ -103,42 +118,134 @@ func lowerSwitchVector[W word.Word[W]](vec BytecodeVector[W], registers split.Al
...
+		bdef   = registers.AllocateNamed(fmt.Sprintf("$b_switch_%d_case_default", switchIndex), util.Some[uint](1))
+	)
+	//
+	codes = append(codes,
+		// one = 1
+		bytecode.LoadConst(onereg, one),
+		// sum = b_0 + ... + b_{n-1}
+		bytecode.AddConst(sumreg, bits, zero),
+		// b_default = one - sum
+		bytecode.NewArith(bytecode.OP_SUB, []descriptor.RegisterId{bdef},
```
DavePearce: I do not understand why we generate `1 - (a + b)` when we can have a single instruction `1 - a - b` which does the same thing?

DavePearce: Using `1 - a - b` is totally safe as it occurs before register splitting.

pkg/zkc/constraints/path_condition.go
```diff
@@ -20,19 +20,23 @@ import (
 )
 
 // TranslateBranchCondition translates a given branch condition within the
-// context of a given state reader.
+// context of a given state reader.  Complement forms over the given one-hot
+// groups are first rewritten into their single-atom equivalents (see
+// rewriteOneHotConditions).
 func TranslateBranchCondition[W vm.Word[W], F field.Element[F], E Expr[F]](p dfa.Path[W],
-	reader RegisterReader[F]) Expr[F] {
+	groups []oneHotGroup, reader RegisterReader[F]) Expr[F] {
```
DavePearce, line 27: This is kind of hideous.  But, I guess, when the branch table optimisation is reworked it can be fixed.

pkg/zkc/vm/internal/transform/lower_switch.go
DavePearce: In fact, looking at the generated IR I see this:

```
...
sub $b_switch_1_case_default = $32-$33                                                                                                                                                                                  
check $b_switch_1_case_default:u1                                                                                                                                                                                       
check $b_switch_1_case_default:u1
```

This strongly suggests to me that `InsertCheckCasts` is doing its job.

pkg/zkc/constraints/one_hot.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// Both substitutions preserve the condition's meaning on any trace satisfying
+// the enclosing vector's constraints (default = 1 - sum of bits, everything a
+// bit), but NOT on arbitrary traces: they must only be applied to conditions
+// of bytecodes within the vector declaring the group.  The rewrite happens
+// here — at the constraint-translation boundary — rather than in the DFA,
+// because the complement form is what lets path conditions cancel where the
+// dispatch's edges rejoin; substituting the default atom there instead would
+// burden every bytecode after the join with an irreducible disjunction.
+func rewriteOneHotConditions(cond dfa.BranchCondition, groups []oneHotGroup) dfa.BranchCondition {
```
DavePearce, line 78: I don't really like what its doing here: basically hacking on top of the `BranchCondition`.  But, we know the branch table optimisation is already quite broken and needs to be reworked.  So when that happens we can try to tidy up this process.

pkg/zkc/vm/internal/bytecode/vector.go
```diff
@@ -482,6 +499,39 @@ func branchTableTransfer[W word.Word[W]](writeMap dfa.Result[dfa.Writes],
...
+			)
+			//
+			for _, c := range code.Cases {
+				var (
+					bid = register.NewId(uint(c.Bit))
+					bit = dfa.NewBranchId(writes.MayAnybeAssigned(bid), bid)
+				)
+				//
+				arcs = append(arcs, dfa.NewTransfer(state.NotEqualsConst(bit, zero), offset+uint(c.Skip)+1))
+				// The fall-through is the syntactic complement of the case
```
DavePearce, line 523: This is a difficult comment to understand, but basically its saying: do it this way so it plays nicely with branch table optimisation.

DavePearce (dismissed):
Overall, this looks good to me.  The only change I am actually requesting is to remove the `CheckCast` added in the `LowerSwitch` transform.  I also would think changing it to generate `1 - a - b` rather than `1 - (a + b)` would make sense.  The issue around the `AllocateName` is fine, and if it bothers me too much at some point I can tweak it.  I also the changes made to override the branch table optimisation are intricate, and difficult to understand.  But, that algorithm is already broken and we can fix that at some point in the future.

pkg/zkc/vm/internal/interpreter/encoding/common.go
```diff
@@ -536,7 +544,7 @@ func RegisterVectorsAsShorts(vecs []RegisterVector) []uint16 {
 }
 
 func init() {
-	if MAX_BYTECODE > OPCODE_MASK {
+	if MAX_BYTECODE > OPCODE_MASK+1 {
```
DavePearce, line 547: Ah, nice.  So we actually still have one more even!

DavePearce (approved):
Yup, nice

PR #2044, opened 2026-07-23
fix: support wide intrinsics in gogen

### LINKED ISSUE #2032: fix: gogen support for `WIDE_XXX` intrinsics
# Summary

Currently, `gogen` fails for tests which use wide intrinsics (e.g. `WIDE_SHL`) arising from fast mode splitting.  An example is `BigNum_04`.

gbotrel:
Fixes #2032.

Adds gogen support for `WIDE_SHL`, `WIDE_SHR`, `WIDE_DIV`, and `WIDE_REM`. Wide shifts use native words; division and remainder use `big.Int`.

Note: first shot used big.Int everywhere and I was worried about perf, this version is a in between perf and verbosity, may revisit if that shows up on profile too much

### DIFF
--- a/pkg/zkc/vm/internal/gogen/emit_field.go
+++ b/pkg/zkc/vm/internal/gogen/emit_field.go
@@ -16,6 +16,7 @@ package gogen
 import (
 	"fmt"
 	"math/big"
+	"strings"
 
 	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/bytecode"
 	"github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/word"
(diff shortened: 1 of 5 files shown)


pkg/zkc/vm/internal/gogen/emit_field.go
```diff
@@ -205,8 +206,67 @@ func (g *generator) emitModPHelpers(c *code) {
...
+	c.line("func wideInt(words []uint64) (out big.Int) {")
+	c.line("if bits.UintSize == 64 {")
+	c.line("limbs := make([]big.Word, len(words))")
+	c.line("for i, word := range words {")
+	c.line("limbs[i] = big.Word(word)")
+	c.line("}")
+	c.line("out.SetBits(limbs)")
+	c.line("return out")
+	c.line("}")
+	c.line("limbs := make([]big.Word, 2*len(words))")
```
gbotrel, line 229: I guess this part is not really needed (i.e. 32 bit machine target)

pkg/zkc/vm/internal/gogen/gen_test.go
```diff
@@ -685,6 +738,20 @@ var diffCases = []diffCase{
 			{"data": {5, 0}}, // division by zero -> error
 		},
 	},
+	{
+		name: "divmod128", // multiword division: WIDE_DIV/REM (fast); wide HINT_DIVISION (lowered)
```
gbotrel, line 724: I find these small tests in gogen more readable but agree that they are a bit diverging from the testing framework

DavePearce (approved):
LGTM

pkg/zkc/vm/internal/gogen/gen_test.go
DavePearce: Yeah, which means they won't be usable, for example, by the prover.

PR #2046, opened 2026-07-23
feat: switch tracing to Uint32

### LINKED ISSUE #2045: feat: switch tracing over to `Uint32`
# Summary

Currently, tracing on `KOALABEAR_16` is using memory particularly poorly as the underlying word is `Uint128`.  In fact, for this field, we should be able to use a `Uint32` word type and this would reduce memory usage by 75%.

DavePearce:
This adds an implementation for Uint32 which is roughly similar to the
existing implementation of Uint64, and switches tracing over to use the
Uint32 implementation.  This should reduce overall memory requirements
for tracing.


letypequividelespoubelles (approved):
LGTM

PR #2047, opened 2026-07-24
fix: bug in lowering division

### LINKED ISSUE #2033: fix: another bug with register splitting multiplication?
# Summary

There is another bit overflow arising for the `ModExp32` benchmark which, presumably, is coming from splitting multiplication.

DavePearce:
This puts through a fix for lowering division which was allocating an
intermediate register with insufficient bits, leading to unexpected
truncation.


letypequividelespoubelles (approved):
lgtm

PR #2048, opened 2026-07-24
feat: support SkipIf with constant operand

### LINKED ISSUE #1741: feat: generalise `vm.SkipIf`
This instruction currently does not allow constants on the right-hand side like the original `micro.SkipIf` did.  This is awkward as it does force us to load constants into registers in certain places which this is not required.  For example, when compiling out bitwise shift operations we could exploit a constant on the right hand side.

DavePearce:


pkg/asm/compiler/branches.go
```diff
@@ -149,6 +166,10 @@ func expandBranchEquality[T any, E Expr[T, E]](p dfa.BranchEquality, reader Regi
 	return expandBranchNonEqualityRegReg(p.Left, p.Right.First())
 }
 
+func isNative[T any, E Expr[T, E]](reg dfa.BranchId, reader RegisterReader[E]) bool {
```
letypequividelespoubelles, line 169: I really don't like it tbh ...

pkg/test/zkc_invalid_test.go
```diff
@@ -308,6 +308,9 @@ func Test_ZkcInvalid_Const_07(t *testing.T) {
 func Test_ZkcInvalid_Const_08(t *testing.T) {
 	checkZkcInvalid(t, "zkc/invalid/const_08")
 }
+func Test_ZkcInvalid_Const_09(t *testing.T) {
```
letypequividelespoubelles, line 311: +1

letypequividelespoubelles, line 311: add a comment to say what it does plz (so no need to open the fil to knwo)

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -212,6 +220,22 @@ func (p *StmtCompiler) compileCondition(pc uint, e Condition, mapping []uint, ta
 	}
 }
 
+func (p *StmtCompiler) normaliseComparison(e *expr.Cmp[symbol.Resolved], lhs, rhs Operand[vm.Uint],
+) (expr.CmpOp, RegisterId, Operand[vm.Uint]) {
+	//
+	if lhs.IsConstant() && rhs.IsConstant() {
+		// This should not be permitted
+		p.errors = append(p.errors, p.srcmaps.SyntaxErrors(e, "branch always taken")...)
```
letypequividelespoubelles: I don't find the error msg very meaningful, I'd prefer "can't compare two constants"

letypequividelespoubelles (dismissed):
LGTM

pkg/asm/compiler/branches.go
DavePearce: Yeah, its not great.  But, its because this component is still legacy.  I will move it over at some point (and actually remove all the legacy stuff), but for now it remains the best option I had.

pkg/test/zkc_invalid_test.go
DavePearce: I don't like writing comments in the test harness.  Its better to just look at the file and/or have a comment there.  Soon the test harness is going to change quite dramatically, and there will be no place for such comments --- they would have to go in the test file itself.

pkg/zkc/compiler/codegen/statement.go
DavePearce: Well, I prefer branch always taken.  But, ok, let's try and find something more readable.  Maybe just `cannot compare constants`.  Yup, that works for me.  I'll make that change now.

DavePearce: Ok, applied and error is now `cannot compare constants`.

pkg/test/zkc_invalid_test.go
DavePearce: The reason for this is the prover is forcing us to move away from this style of writing tests.  Instead, all test information will be embedded in the test file itself (e.g. `skip=true`, `constraints=false`, etc).  That way, when the prover reads the test, its knows whether or not to try and run and, if so, what it can expect (e.g. `gogen` does work or something).

letypequividelespoubelles (approved):
lgtm

PR #2050, opened 2026-07-24
feat: add range constraint for ROM/WOM/RAM modules

### LINKED ISSUE #2049: add range check for memoty module

letypequividelespoubelles:


DavePearce (approved):
LGTM --- pretty straightforward.  I think at some point we'll need to tidy up `translator.go`, but that is for another day.

PR #2052, opened 2026-07-24
chore: panic instead of log.warn for untouched registers

### LINKED ISSUE #2051: chore: from log.warn to error for untouched register

letypequividelespoubelles:


pkg/cmd/zkc/compile.go
```diff
@@ -121,15 +121,15 @@ func writeArtifacts[F field.Element[F]](filename string, build BuildConfig, arti
...
+// constraint or lookup.  If any such register is encountered, this fires errors which identify the enclosing
+// modules and registers.
 func validateArtifacts[F field.Element[F]](field field.Config, artifacts BuildArtifacts, config CompileConfig) {
 	// Generate AIR representation
 	air := constraints.GenerateAirConstraints[vm.Uint, F](artifacts.ir, field, config.build.config.GetMaxStaticDepth())
 	// Perform validation check
 	if errs := constraints.Validate(air); len(errs) > 0 {
 		for _, err := range errs {
-			log.Warn(err)
+			fmt.Printf("panic: untouched register: %v\n", err)
```
DavePearce: @letypequividelespoubelles You should actually panic here.

DavePearce (changes requested):
You need to actually `panic`.

DavePearce (approved):
Yeah, that looks better!

PR #2061, opened 2026-07-28
feat: support space-efficient trace format

### LINKED ISSUE #2059: feat: compact trace representation
# Summary

The goal of this feature is to offer a compact trace representation, such that the memory footprint whilst tracing is reduced.  The idea for how this works is to revert to a column-oriented format and reuse existing efficient array forms (e.g. for `bool`, `uint16`, etc).

DavePearce:


pkg/rtrace/compact.go
```diff
@@ -0,0 +1,203 @@
...
+func (p *CompactModule[T]) Name() string {
+	return p.name
+}
+
+// Descriptor returns the registers describing this module.
+func (p *CompactModule[T]) Descriptor() iter.Iterator[Register] {
+	return iter.NewArrayIterator(p.registers)
+}
+
+// RegisterAt returns the limb with the given index.
```
letypequividelespoubelles: this is the column index right ? So better have the comment `returns the register with` 
Don't mix limb / register terminology if that's the same thing

pkg/rtrace/compact.go
```diff
@@ -0,0 +1,203 @@
...
+func (p *CompactModule[T]) RegisterAt(index uint) Register {
+	return p.registers[index]
+}
+
+// Row returns a specific row within this trace module.
+func (p *CompactModule[T]) Row(id uint) Row[T] {
+	return &compactRow[T]{p.columns, id}
+}
+
+// Height returns the height of this module, meaning the number of assigned
```
letypequividelespoubelles, line 109: same, we have to harmonize the terminology, sometime we speak of height, sometimes of depth for the same thing: the nb of rows

pkg/rtrace/compact.go
```diff
@@ -0,0 +1,203 @@
...
+// rows.
+func (p *CompactModule[T]) Height() uint {
+	if len(p.columns) == 0 {
+		return 0
+	}
+	//
+	return p.columns[0].Len()
+}
+
+// Width returns the number of columns in this module.
```
letypequividelespoubelles, line 119: columns => register ?

pkg/rtrace/compact.go
```diff
@@ -0,0 +1,203 @@
...
+	row uint
+}
+
+// Get the value in a given column of this row.
+func (p compactRow[T]) Get(column uint) T {
+	return p.columns[column].Get(p.row)
+}
+
+// Returns the number of columns in this row.
+func (p compactRow[T]) Width() uint {
```
letypequividelespoubelles: tbh naming are confusing me with:

```
func (p *CompactModule[T]) Width() uint {
	return uint(len(p.registers))
}
```

we have few lines before.

make distinction column / register / limb more clear

pkg/rtrace/compact.go
```diff
@@ -0,0 +1,203 @@
...
+	return uint(len(p.columns))
+}
+
+func allocArray[F field.Element[F]](bitwidth uint, height uint) narray.MutArray[F] {
+	var (
+		zero  F
+		width = zero.Modulus().BitLen()
```
letypequividelespoubelles, line 170: names are confusing here: here width == moduleBitWidth ? and bitwidth == registerBitWidth ?

pkg/rtrace/compact.go
```diff
@@ -0,0 +1,203 @@
...
+		return narray.NewConstantArray(height, 0, zero)
+	case bitwidth == 1:
+		return narray.NewBitArray[F](height)
+	case bitwidth <= 8 && width >= 8:
+		return narray.NewSmallArray[uint8, F](height, bitwidth)
+	case bitwidth <= 16 && width >= 16:
+		return narray.NewSmallArray[uint16, F](height, bitwidth)
+	case bitwidth <= 32 && width >= 32:
+		return narray.NewSmallArray[uint32, F](height, bitwidth)
+	case bitwidth <= 64 && width >= 64:
```
letypequividelespoubelles, line 184: is this even reachanble for KoalaBear ?

letypequividelespoubelles (changes requested):
LGTM, but some naming are confusing, please make it more explicit

pkg/rtrace/compact.go
DavePearce: Hmmmm, that's claude did that.  Hmmmm, it happened because I refactored that's why.  Yup will fix.

pkg/rtrace/compact.go
DavePearce: Hmmmmm, in `go-corset` I don't think I've ever used depth.  The term `Height` has existed for a long time in the `trace.Module` interface, for example.

pkg/rtrace/compact.go
DavePearce: No, its not.  But, we are designing this to support other fields, such as Goldilocks, etc

pkg/rtrace/compact.go
DavePearce: Done

pkg/rtrace/compact.go
DavePearce: Yeah, ok.  In this context, they are kind of interchangeable.  But, yeah, seems reasonable.

pkg/rtrace/compact.go
DavePearce: Ok, there should be no "limb" at all --- that is an artifact from how it looked before which I removed.  So, I did a search and fixed all occurrences of limb / Limb.

DavePearce: Have now reworked the terminology quite a lot.

pkg/rtrace/compact.go
DavePearce: have reworked the terminology

pkg/cmd/zkc/flags.go
```diff
@@ -111,10 +111,10 @@ func checkFlags(cmd *cobra.Command, checks FlagChecks) {
 	// Some sanity check on some flag values:
 	// Note: range checks require static tables of size at least 2.
 	// no more the case once https://github.com/LFDT-Lineth/zkc/issues/1910 implemented
-	if cmd.Flags().Changed("max-static-depth") {
-		maxStaticDepth := GetUint(cmd, "max-static-depth")
-		if maxStaticDepth <= 2 {
-			fmt.Printf("error: \"--%s\" must be greater than 2\n", "max-static-depth")
+	if cmd.Flags().Changed("max-static-height") {
+		maxStaticHeight := GetUint(cmd, "max-static-height")
+		if maxStaticHeight <= 2 {
```
letypequividelespoubelles, line 116: actually I don't like this (I know it's mine, not yours :) )
It's a reminiscence of my first implem of range constraints. But actually we shouldn't have this, but a compile check that the max declared static table in .zkc is >= this value. 

I'll open an issue for it.

letypequividelespoubelles (approved):
LGTM !

PR #2070, opened 2026-07-29
perf: optimize check against -1

### LINKED ISSUE #2063: optimize check against -1
Consider the following:
```
fn test(x:u64) -> (r:u16) {
    r = (x == 0xFFFFFFFFFFFFFFFF) ? 1 : 0
}
```
It generates this degree 10 consraint
```
(vanish test:pc0#4 (
   *
   $ret
   (- 1 (* (- x'3 65535) "(inv (- test:x'3 65535))"))
   (- 1 (* (- x'2 65535) "(inv (- test:x'2 65535))"))
   (- 1 (* (- x'1 65535) "(inv (- test:x'1 65535))"))
   (- 1 (* (- x'0 65535) "(inv (- test:x'0 65535))"))
   (- $2 1)))
```
of which 8 degrees are coming from comparaisons.

It is currently the culprit for generating highest degree (15) for r5-zkc interpretor.

letypequividelespoubelles:


pkg/zkc/constraints/call_and_memory_lookup.go
letypequividelespoubelles, line 1: this is just cleaning

DavePearce (commented):
One observation I had is that, whilst this does seem to reduce overall complexity, it can lead to constraints with higher degree.  My Example was this:

```
fn f(x:u64) -> (r:u1) {
   if x == 0xffff_ffff_ffff_ffff {
     r = 1
   } else {
     r = 0
   }
}
```

Which previously generate constraints with max degree 3, but now generates with max degree 4.  Maybe this is related to #2071 ?  The constraint reported with degree `4` is this one:

```
(vanish f:pc0#0 (
   *
   $ret
   $2
   (- 1 (
      *
      (+ (- 65535 x'3) (- 65535 x'2) (- 65535 x'1) (- 65535 x'0))
      "(inv (+ (- 65535 f:x'3) (- 65535 f:x'2) (- 65535 f:x'1) (- 65535 f:x'0)))"))))
```

pkg/zkc/constraints/call_and_memory_lookup.go
```diff
@@ -164,71 +164,27 @@ outer:
...
-	// In a multi-line caller the call only fires on rows executing its line, so
-	// fold the line's PC selector (IS_PC_k != 0) into the condition as an
-	// extra single-bit atom.  An atomic caller has no line selectors, so its
-	// gating is the branch condition alone.
+	// Position register gating the rows of this access: the line's IS_PC_k
+	// selector for a multi-line function, or $ret for an atomic one (defining
+	// the non-padding region).
+	var position register.Id
+	//
 	if len(pcSelectors) != 0 {
```
DavePearce, line 172: Hmmmm, so what is this lot doing?  Seems nice to remove a good of that code.

DavePearce: Seems like basically this is just dropping the call to `singleBitGuard()`.  Does that mean the optimisation no longer applies?

DavePearce (approved):
Overall, LGTM.  I was a bit confused why check to `singleBitGuard()` was removed, and whether this means an optimisation is lost.  See my comment also about how this can lead to an *increase* in degree.

letypequividelespoubelles:
> One observation I had is that, whilst this does seem to reduce overall complexity, it can lead to constraints with higher degree. My Example was this:
> 
> ```
> fn f(x:u64) -> (r:u1) {
>    if x == 0xffff_ffff_ffff_ffff {
>      r = 1
>    } else {
>      r = 0
>    }
> }
> ```
> 
> Which previously generate constraints with max degree 3, but now generates with max degree 4. Maybe this is related to #2071 ? The constraint reported with degree `4` is this one:
> 
> ```
> (vanish f:pc0#0 (
>    *
>    $ret
>    $2
>    (- 1 (
>       *
>       (+ (- 65535 x'3) (- 65535 x'2) (- 65535 x'1) (- 65535 x'0))
>       "(inv (+ (- 65535 f:x'3) (- 65535 f:x'2) (- 65535 f:x'1) (- 65535 f:x'0)))"))))
> ```

I don't understand: 
for: 

```
fn test(x:u128) -> (r:u16) {
    r = (x == 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF) ? 1:0
}
```

with this PR:
image


because of this degree 4 consrraint:
image


vs without: 
image

because of this degree 18: 
image


so for `test`we go from maxDegree 18 to maxDegree 4

pkg/zkc/constraints/call_and_memory_lookup.go
letypequividelespoubelles: it's just that now we reusing the mechanism from branchCondition to simplify b == u1 const. Before it was creating degree 2 as the constant was embedded in a register, so b == u1 const wasn't a bool anymore. 

So now, reove now unused code thanks to const in skip if

PR #2074, opened 2026-07-29
feat: add padding for static tables

### LINKED ISSUE #2064: fix: padding for static reference tables
# Summary

It would appear that static reference tables also need to be padded up to the next power-of-two.  The purpose of this feature is simply to implement this change.  Since we can assume that a static reference table is never empty, then presumably we can pad with data from e.g. the  first row.  Otherwise, the other obvious approach would to add a dedicate `$ret` line.

## Example

Minimal example to demonstrate:

```
static mem(addr:u32) -> (b:u8) {
  123,254,10,11,234
}
```

This then generates the following `MIR` output:
```
(module mem :static)

(inputs
   (addr'1 u16 0x0)
   (addr'0 u16 0x0)
)

(outputs
   (b u8 0x0)
)

(contents
   (0x0, 0x0, 0x7b),
   (0x0, 0x1, 0xfe),
   (0x0, 0x2, 0xa),
   (0x0, 0x3, 0xb),
   (0x0, 0x4, 0xea)
)

...
```

letypequividelespoubelles:


pkg/cmd/zkc/compile.go
```diff
@@ -120,18 +120,30 @@ func writeArtifacts[F field.Element[F]](filename string, build BuildConfig, arti
...
+// constraint or lookup.
+// - static tables are of power of two depth
 func validateArtifacts[F field.Element[F]](field field.Config, artifacts BuildArtifacts, config CompileConfig) {
 	// Generate AIR representation
 	air := constraints.GenerateAirConstraints[vm.Uint, F](artifacts.ir, field, config.build.config.GetMaxStaticDepth())
-	// Perform validation check
-	if errs := constraints.Validate(air); len(errs) > 0 {
-		for _, err := range errs {
+	// validate that all registers are referenced in at least one vanishing constraint or lookup
+	untouchedRegister := constraints.Validate(air)
```
OlivierBBB: Is `untouchedRegister` still an error ? Tiny nitpick: the name should be plural.

pkg/zkc/constraints/util.go
```diff
@@ -109,6 +111,22 @@ func foldContents[F field.Element[F]](inputs, outputs []register.Register, conte
...
+// power-of-two height, by duplicating the last row.
+func padStaticTables[F field.Element[F]](rows [][]F) [][]F {
+	if len(rows) == 0 {
+		panic("A static table can't be declared empty")
+	}
+	//
+	var target = pow.NextPowerOfTwo(uint(len(rows)))
+	//
+	for len(rows) < int(target) {
+		rows = append(rows, slices.Clone(rows[len(rows)-1]))
```
OlivierBBB, line 124: Is the rationale for padding with the final row rather than a trivial row at the start to prevent undesirable matches in the lookup ?

OlivierBBB (dismissed):
LGTM

pkg/zkc/constraints/util.go
letypequividelespoubelles: rationale is: if we pad with starting value, it adds an offset for all the value we had at the begining. Plus in case we need to humanly read the static tables, it'll start with duplicate. So imho better to add padding at the end. In theory it has 0 impact, just for readability.

letypequividelespoubelles: (of course you can't pad with a 0 row, you need to pad with an already present value, if you pad with 0, then you need to add a $ret colun, and add this column as the target selector for lookups, etc)

pkg/cmd/zkc/compile.go
letypequividelespoubelles: yes, name updated.

pkg/cmd/zkc/compile.go
```diff
@@ -120,18 +120,30 @@ func writeArtifacts[F field.Element[F]](filename string, build BuildConfig, arti
...
+	// validate that all registers are referenced in at least one vanishing constraint or lookup
+	untouchedRegisterErrs := constraints.Validate(air)
+	if len(untouchedRegisterErrs) > 0 {
+		for _, err := range untouchedRegisterErrs {
 			log.Errorf("untouched register: %v", err)
 		}
-		//
+	}
+	// validate static tables size
+	staticTablesSizeErrs := constraints.ValidateStaticTablesSize(air)
```
DavePearce: Can you move the check in `ValidateStaticTablesSize()` into `Validate()` please? We only need one function for validation.

pkg/zkc/constraints/util.go
DavePearce: Yeah, padding at the back seems fine to me.

DavePearce (approved):

PR #2078, opened 2026-07-30
feat: add Module.IsPublicOutput / IsPrivateOutput

### LINKED ISSUE #1946: feat: distinguish `pub input` from `pub output` in `schema.Module`
# Summary

The goal here is to allow one to distinguish the program inputs from the program outputs at the constraints level.   That is, when looking at a `schema.Module` instance where `IsPublic()` holds we should additionally be able to check whether it is intended to be `IsOutput()` as well.

DavePearce:
This adds functions to schema/Module for identifying modules which correspond to public and private outputs.

### DIFF
--- a/pkg/asm/io/function.go
+++ b/pkg/asm/io/function.go
@@ -120,12 +120,19 @@ func (p *Function[T]) Code() []T {
 	return p.code
 }
 
-// IsPublic determines whether or not this is an "externally visible" function
-// or not.  The differences between internal and external functions is small.
-// Specifically, internal functions are not visible in the generated trace
-// interface; likewise, they are hidden by default in the inspector.
-func (p *Function[T]) IsPublic() bool {
+// IsPublicOutput determines whether or not this is an "externally visible"
+// function (e.g. a public ZkC output memory) or not.  The differences between
+// internal and external functions is small.  Specifically, internal functions
+// are not visible in the generated trace interface; likewise, they are hidden
+// by default in the inspector.
+func (p *Function[T]) IsPublicOutput() bool {
 	return p.public
+}
+
+// IsPrivateOutput indicates whether or not this module represents a private
+// output.  At this time, functions can never be private outputs.
+func (p *Function[T]) IsPrivateOutput() bool {
+	return false
 }
 
 // IsSynthetic modules are generated during compilation, rather than being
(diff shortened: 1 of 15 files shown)


pkg/asm/io/function.go
```diff
@@ -120,14 +120,21 @@ func (p *Function[T]) Code() []T {
...
-// or not.  The differences between internal and external functions is small.
-// Specifically, internal functions are not visible in the generated trace
-// interface; likewise, they are hidden by default in the inspector.
-func (p *Function[T]) IsPublic() bool {
+// IsPublicOutput determines whether or not this is an "externally visible"
+// function (e.g. a public ZkC output memory) or not.  The differences between
+// internal and external functions is small.  Specifically, internal functions
+// are not visible in the generated trace interface; likewise, they are hidden
+// by default in the inspector.
+func (p *Function[T]) IsPublicOutput() bool {
```
letypequividelespoubelles, line 128: don't we always isOutput == (isPublicOutput or isPrivateOutput) ? 
If yes, I we should have only one bool if not I find the naming confusing

pkg/schema/module.go
```diff
@@ -31,8 +31,13 @@ type ModuleView interface {
 	register.Map
 	// Module name
 	Name() module.Name
-	// IsPublic indicates whether or not this module is externally visible.
-	IsPublic() bool
+	// IsPublicOutput indicates whether or not this module represents a public
+	// output (e.g. a public ZkC output memory) and, hence, is externally
+	// visible.
+	IsPublicOutput() bool
```
letypequividelespoubelles, line 37: don't we always isOutput == (isPublicOutput or isPrivateOutput) ? 
If yes, I we should have only one bool if not I find the naming confusing

Will we need input as well ? 

If yes, I'd suggest to have 3 bool; input (ROM ?)  / output (WOM ? ) + public or private.

letypequividelespoubelles (approved):
LGTM with small concern on terminology

pkg/asm/io/function.go
DavePearce: I can't understand this

pkg/schema/module.go
DavePearce: Well, technically I do not believe we need e.g. `IsPublicInput()` at this time.  However, yes, I could have stuck with `IsPublic()` and `IsOutput()` and `IsInput()` and, to be honest, that was my original plan.  BUT, I decided against adding `IsPublicInput()` ... because I am concerned it will cause confusion.

PR #2079, opened 2026-07-30
feat: timestamp threading for read-write memories (1/3)

OlivierBBB:
Part 1 of 3 of the RAM timestamps feature (#1807), following the review of #2030: this PR threads the stamps through the machine; RAM module constraints + tracing (2/3) and the caller→memory lookups consuming the stamp operand (3/3) follow as stacked PRs.

## What it does

Every function declaring a read-write memory effect gains a stamp-in input and stamp-out output register per accessed memory (stamps first, named `<mem>$stamp` / `<mem>$stamp_out`); calls to effectful callees forward them, and every memory access records its timestamp in the new `ReadWrite.Stamp` operand — the k-th access executed carries `stamp_in + k`. The threaded stamp exists for tracing / constraint generation only (the run-time memory keeps its own clock), so the transform runs on the constraint path only; fast mode is untouched.

## How

The threading is **offset-based**: the transform tracks, at every program point, which register holds the current stamp plus a constant offset. A statement performing several accesses — read-feeds-write (`ram[x] = ram[x] + 1`), two reads in one expression or condition, nested addressing (`ram[ram[x]] = ram[y]`) — gives each access a distinct timestamp without ever writing a stamp register twice on one path. The canonical stamp register is written at most once per executed path through a row, so **one-line functions stay one-line** (there is no vectorise barrier), and the common single-access function costs exactly one added instruction:

```
fn f<ram>(x:u16) -> (r:u8) { r = ram[x] }
  ⇓
read r = ram[ram$stamp; x]
add ram$stamp_out = ram$stamp + 1
```

The transform runs **after vectorisation** (a vector is then genuinely one trace row; the walk understands the vectoriser's skip-based control flow, including calls inside skip regions and dispatch tables) and **before register splitting** (stamp registers and their arithmetic split into limbs and are range-checked for free).

## Review feedback addressed (from #2030)

- generated code near-optimal (the example above is the reviewed-thread's stated optimum);
- multi-access statements (`dinc` / `dset` / `eq` and nested forms) all work — covered by the new `ram_01..08` fixtures;
- no vectorise barrier; multiple accesses (same or distinct memories) on one row are fine;
- memory effects are gob-serialised and preserved by every transform (enforced via the `descriptor.NewFunction` signature);
- `NewMemRead` / `NewMemWrite` take the stamp as an optional constructor operand, used by every transform that rebuilds an access;
- stamp-out named `$stamp_out` (no collision with the `'k` limb suffix);
- all bench tests pass (Blake / Sort / Poseidon / Keccakf), nothing skipped.

## Testing

New end-to-end fixtures `testdata/zkc/unit/ram_01..08` (execution + gogen + constraint paths, `.accepts`/`.rejects`). Full `Test_ZkcUnit|Test_ZkcMixed|Test_ZkcInvalid`, the full bench suite, the racer target, and the gogen goldens are green; lint clean.


pkg/zkc/vm/internal/bytecode/bytecode.go
```diff
@@ -402,9 +404,24 @@ func SubVecConst[W word.Word[W]](targets []RegisterId, sources []RegisterId, con
...
+// singleStamp unwraps the optional variadic stamp operand of NewMemRead /
+// NewMemWrite, accepting either no stamp or exactly one.
+func singleStamp(stamp [][]RegisterId) []RegisterId {
+	switch len(stamp) {
+	case 0:
+		return nil
+	case 1:
+		return stamp[0]
+	default:
+		panic("at most one stamp operand allowed")
```
DavePearce, line 427: Does this not seem wierd?  Its preventing have multiple stamp lanes.

pkg/zkc/vm/internal/transform/inline_functions.go
```diff
@@ -515,11 +517,22 @@ func substituteRegisters[W word.Word[W]](insn Bytecode[W], sub []bytecode.Regist
...
 		return insn
+	case *bytecode.Jmp[W], *bytecode.Ret[W]:
+		// Register-free; present only when substituting a full function body.
+		return insn
 	case *bytecode.SkipIf[W]:
 		return &bytecode.SkipIf[W]{Op: insn.Op, Skip: insn.Skip,
 			Left: substituteRegisterVector(insn.Left, sub), Right: substituteOperandVector(insn.Right, sub)}
 	case *bytecode.Switch[W]:
 		return bytecode.MultiwaySkip(substituteId(insn.Source, sub), insn.Cases)
+	case *bytecode.Dispatch[W]:
```
DavePearce, line 528: That's interesting --- this case must have been missing from when that feature was merged!

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+// global default.
+const stampWidth uint = 32
+
+// ThreadTimestamps threads a per-memory timestamp through every function which
+// declares a read-write memory effect.  For each such function and read-write
+// memory M it may access, it:
+//
+//   - adds an input register "M$stamp" and an output register "M$stamp_out",
+//     placed first in the inputs / outputs.  The entry function "main" takes
+//     no parameters, so it gets no stamp in/out and instead counts from ONE
```
DavePearce, line 39: Ok, that's interesting we starting timestamps from `1` ... I wasn't expecting that.  I guess you specifically requested that?  Seems fine eitherway.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+//     (timestamp zero is reserved for the initial state of an untouched cell);
+//   - forwards the stamp across every call to an effectful callee (prepended
+//     to the call's arguments and returns);
+//   - gives every access a distinct timestamp: the k-th access executed
+//     carries stamp_in + k, recorded in the access's Stamp operand.
+//
+// The threading is offset-based: rather than incrementing a working register
```
DavePearce: When it says "offset-based", I think I would say "version based".  Specifically, my approach would be based on a variation of SSA form.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+// rwMemoryEffects returns the module ids of the read-write memories declared as
+// effects of the given function, preserving declaration order.  Effects naming
+// a read-only / write-once memory (which need no timestamp) are dropped.
+func rwMemoryEffects[W word.Word[W]](mods []descriptor.Module[W],
+	fn *descriptor.Function[W]) []descriptor.ModuleId {
+	//
+	var out []descriptor.ModuleId
+	//
+	for _, id := range fn.Effects() {
+		if mem, ok := mods[id].(*descriptor.Memory[W]); ok && mem.IsReadWrite() {
```
DavePearce: This seems odd to me because (as it currently stands) a memory effect can given for a ReadWrite memory.  It is a syntax error to use a ROM/WOM as a memory effect.  So, I guess this function can be dropped (at least for now).

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+			out = append(out, id)
+		}
+	}
+	//
+	return out
+}
+
+// stampState records, symbolically, where the current timestamp of one memory
```
DavePearce: What's missing here is a note to clarify that the temporary registers introduced for a RAM are always allocated consecutively.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+	base util.Option[bytecode.RegisterId]
+	off  uint64
+}
+
+// equals reports whether two states denote the same symbolic value.
+func (s stampState) equals(o stampState) bool {
+	return s.base == o.base && s.off == o.off
+}
+
+// bump returns the state advanced by one access.
```
DavePearce: Yes, this looks much like the SSA-based versioning technique.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+	// stamp-out output or, for main, a lazily-allocated computed register.
+	// Every row except the entry row is entered with the current stamp held
+	// there, at offset zero.
+	canonical map[descriptor.ModuleId]bytecode.RegisterId
+	isMain    bool
+}
+
+// threadFunction rewrites a single function to thread a timestamp for each of
+// the given read-write memory effects.
+func threadFunction[W word.Word[W]](mods []descriptor.Module[W], fn *descriptor.Function[W],
```
DavePearce, line 142: Overall, this function is pretty ugly --- I think we will want to refactor it.  I will try to provide some suggestions.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+			effects:   effects,
+			stampIn:   map[descriptor.ModuleId]bytecode.RegisterId{},
+			canonical: map[descriptor.ModuleId]bytecode.RegisterId{},
+			isMain:    isMain,
+		}
+	)
+	// New register layout, stamps first: [stamp-ins, old inputs, stamp-outs,
+	// old outputs, old computed].  main takes no parameters, so it gets no
+	// stamp in/out; its canonical registers are allocated lazily below.
+	if !isMain {
```
DavePearce: Initialising `newRegs` can be factored into an `initNewRegs()` function or similar.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+			newRegs = append(newRegs, descriptor.NewRegister(register.OUTPUT_REGISTER, stampOutName(mods, e),
+				util.Some(stampWidth), padding))
+		}
+	}
+	//
+	newRegs = append(newRegs, oldRegs[ni:]...)
+	// Old->new register id remap: inserting k stamp inputs shifts old inputs
+	// by k; inserting k stamp outputs shifts old outputs and computed by a
+	// further k.  main inserts none, so the map is the identity there.
+	sub := make([]bytecode.RegisterId, len(oldRegs))
```
DavePearce: The initialisation of `sub` and then its application to refactor the entire function can be factored into its own function.  Something like "remapFunction()`".  This function would call `initNewRegs()` function I suggested above.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+					return []Bytecode[W]{bytecode.Jump[W](jmp.Target + 1)}
+				}
+				//
+				return []Bytecode[W]{insn}
+			})
+		}
+	}
+	// Thread each row.  Every row after the first is entered with canonical
+	// stamps (the walk materialises them at each row boundary).
+	for vi := range nvecs {
```
DavePearce: I would probably put this loop, and the `if jumpTarges(nvecs)` condition above into its own function, called something like `threadFunction()`.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+			entry[e] = stampState{util.None[bytecode.RegisterId](), 1}
+		} else {
+			entry[e] = stampState{util.Some(t.stampIn[e]), 0}
+		}
+	}
+	// Every jump lands on a row entered with canonical stamps; if the entry
+	// row is itself a jump target (e.g. a leading loop header), the entry
+	// state is materialised on a preamble row of its own and every jump
+	// target shifted by one.
+	if jumpTargets(nvecs).Contains(0) {
```
DavePearce: I would suggest creating a function with a name like `isXXX()` which has a better comment, and for which we can write `if isXXX(nvecs)` to try and make this more readable.  I would then factor the body of the condition into its own function, with a name like e.g. `applyXXX`.  Let me try and figure out what `XXX` is here, as I'm not sure what this bit is doing.

DavePearce: Hmmmm, is the right name something like `entryIsJumpTarget(nvecs)` perhaps?  So we can write `if entryIsJumpTarget(nvecs) { assignEntryStamps(...) }` or something like that.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+			}
+			//
+			continue
+		}
+		//
+		if !live {
+			continue
+		}
+		//
+		switch insn := insns[i].(type) {
```
DavePearce: This can go into its own function, something like `threadBytecode` or `threadInstruction`.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+		if inserts[i] == nil {
+			inserts[i] = &rowInserts[W]{}
+		}
+		//
+		return inserts[i]
+	}
+	//
+	for i := 0; i <= n; i++ {
+		// Merge skip-carried states into the fall-through state.
+		states, live = t.mergeAt(insns, i, states, live, pending, at)
```
DavePearce: So this function is whether things start to get quite messy in terms of understanding what its doing.  It does look a bit like a data-flow analysis now, since we are merging states --- which roughly I would expect.  However, the meaning of the state that its merging is not clear to me.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+				states = t.threadBranchCase(insns, i, ci, int(c.Skip), states, pending, at)
+			}
+		}
+	}
+	// Rebuild the row, recomputing intra-row skip amounts around insertions.
+	return rebuildVector(insns, inserts, replace)
+}
+
+// mergeAt reconciles the fall-through state with any skip-carried states at
+// position i.  An effect needs equalising when its incoming states disagree or
```
DavePearce: So, in SSA terminology, "equalising" is like inserting Phi node (if I understand correctly).

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+	//
+	at := func(i int) *rowInserts[W] {
+		if inserts[i] == nil {
+			inserts[i] = &rowInserts[W]{}
+		}
+		//
+		return inserts[i]
+	}
+	//
+	for i := 0; i <= n; i++ {
```
DavePearce, line 437: This is a tricky idea, but I'm interested if we can do something here to make this process easier to understand.  Specifically, this is some kind of implementation of a "dataflow analysis".  In fact, we already have framwork in `zkc` for implement "dataflow analyses" which is used in other places.  The framework lives in `pkg/asm/io/micro/dfa` and I think it would be good if we can ask Claude to try and reuse this framework, instead of having a bespoke loop here.  It may come back and say "I can't use that framework for this reason", or it may come back and say "I can use the framework, but only if I completely rework it" --- which I don't think we should agree to.

I might try and ask it now whether it thinks this is theoretically possible or not.

DavePearce: A follow on to this is that it might make sense to precompute the register versioning information (what it calls "offset") before applying the rewrites.  The idea would be: compute register versioning using existing DFA framework, then allocate registers as required for the versions and, finally, apply rewrites for all instructions using this information.  The main complexity would then live in the initial versioning computation, and what came after that would be straightforward.

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+
+// mergeAt reconciles the fall-through state with any skip-carried states at
+// position i.  An effect needs equalising when its incoming states disagree or
+// the merged state is not a plain register.  Each incoming edge materialises
+// the merge register at a position executed exactly on that edge: inline for
+// the fall-through, at the source's preEdge for an unconditional edge, in an
+// edge block at the landing for a conditional edge.  At an exit (return, jump,
+// end of row) the merge register is the canonical stamp itself, so the exit
+// needs no further materialisation on any path.
+func (t *threader[W]) mergeAt(insns []Bytecode[W], i int, states map[descriptor.ModuleId]stampState,
```
DavePearce: Mon Dieu!

pkg/zkc/vm/internal/transform/thread_timestamps.go
```diff
@@ -0,0 +1,909 @@
...
+		case *bytecode.Fail[W]:
+			live = false
+		case *bytecode.Skip[W]:
+			// Unconditional forward edge: the state flows to the landing; the
+			// position before the skip can host edge-only instructions.
+			t.recordEdge(pending, i, 0, int(insn.Skip), states, uncondEdge)
+
+			live = false
+		case *bytecode.SkipIf[W]:
+			states = t.threadBranch(insns, i, int(insn.Skip), states, pending, at)
```
DavePearce: Yeah, this does look relatively similar to how the existing DFA framework works.

pkg/zkc/vm/internal/transform/thread_timestamps.go
DavePearce: This is my prompt:

> The threadVector function in the file thread_timestamps.go implements something along the lines of a dataflow analysis.  Since there is an existing framework for describing dataflow analyses in pkg/asm/io/micro/dfa, I am wondering how you would convert this function to use that framework?  In particular, I am imaging that it would make sense to first compute the register versioning information first using the framework, followed by a sweep which rewrites each bytecode using that information.

Let's see what it comes back with.

DavePearce (changes requested):
ok, a bunch of comments on this one.  Overall, the transformed code looks pretty code, but the actual implementation needs refactoring.  Claude believes we can refactor `threadVector` to use the existing DFA framework (see my comments), and I'm going to try that now to see if it improves things.

pkg/zkc/vm/internal/transform/thread_timestamps.go
OlivierBBB: I did. To make the comparison `TIMESTAMP_READ < TIMESTAMP_WRITTEN` universally true in the RAM module during the `EXEC` phase.

During the `FINL` phase the condition is relaxed to `TIMESTAMP_READ ≥ TIMESTAMP_WRITTEN` and `TIMESTAMP_WRITTEN = 0` (which corresponds to intialization). The switch to `≥` (rather than `>`) is due to us imposing that the first address be the 0 address: we have no guarantee that address 0 will indeed have been touched.

This could be strengthened to `>` if we didn't require the first address in the `FINL` phase to be 0. Which actually makes sense.

pkg/zkc/vm/internal/transform/thread_timestamps.go
OlivierBBB: FYI: SSA is also what Fable had in mind when we were discussing the plan.

DavePearce (dismissed):
LGTM

DavePearce (approved):
LGTM

Issue #2068, opened 2026-07-28
feat: log the full state of memory for the FINL phase of RAM


OlivierBBB:
One trace segment (e.g. the last segment) will have to perform memory finalization. Initialization/Finalization requires full knowledge of the final `[]TimestampedCell` memory contents. This information isn't recorded at the present time. It could be included in a "finalization checkpoint".

PR #2081, opened 2026-07-30
feat: caller lookups for read-write memory accesses (3/3)

OlivierBBB:
Part 3 of 3 of the RAM timestamps feature (#1807), stacked on #2080 (which stacks on #2079); the diff shown here is only this PR's changes.

## What it does

Every read-write memory access now emits a lookup tying it to a row of that memory's table, through the same per-site machinery as calls and access-once memories: the source tuple carries the accessor's address, data and threaded timestamp registers — gated by the shared line/path selector — and maps onto the table's `ADDRESS`, `VALUE_WRITTEN` and `TIMESTAMP_WRITTEN` columns. Multiple accesses on one row each get their own lookup. A read's data registers are its outputs (the value read back), which the table exposes in `VALUE_WRITTEN` too, since a read row writes back what it found (`VALUE_READ == VALUE_WRITTEN`).

The access's read/write kind is pinned by the **target-side filter** rather than a tuple element (MIR lookup tuples carry register accesses only): the table gains two fully-determined selector columns, `EXEC_WRITE = EXEC · IS_WRITE` and `EXEC_READ` (via `EXEC_WRITE + EXEC_READ == EXEC`), filled by the observer. A write site's lookup targets the table filtered by `EXEC_WRITE`, a read site's by `EXEC_READ`.

With this, the guest-facing columns of every access row are pinned by the caller, and the row-local constraints hold end-to-end. The remaining gap — that `VALUE_READ` genuinely returns the *last* value written to the address — is the offline memory-checking bus, tracked in #2068 together with the finalization rows.

## Testing

The `ram_01..10` fixtures and the RAM-heavy benches (blake / sort / keccakf / poseidon) now run with the complete chain live — threading → table → lookups — across the full `Test_ZkcUnit|Test_ZkcMixed|Test_ZkcInvalid` and bench suites; lint clean.


pkg/zkc/constraints/call_and_memory_lookup.go
```diff
@@ -122,17 +127,11 @@ func groupLookupsByCondition[W vm.Word[W]](codes []vm.Bytecode[W], branchTable d
 	//
 outer:
 	for cc, code := range codes {
-		switch c := code.(type) {
+		switch code.(type) {
 		case *vm.BytecodeCall[W]:
 			// always emits a lookup
 		case *vm.BytecodeReadWrite[W]:
```
letypequividelespoubelles, line 133: cleaning: merge the two cases: vm.BytecodeCall and m.BytecodeReadWrite

PR #2089, opened 2026-07-31
2077 feat create space for more bytecode encodings

### LINKED ISSUE #2077: feat: create space for more bytecode encodings
# Summary

Currently, we are at the limit of space for adding new instruction encodings.  Specifically, every opcode is a single `u8` where bit `7` (i.e. the most significant bit) indicates whether a breakpoint is present, and bit `6` indicates whether or not this is a `WIDE` instruction.  However, to support more instructions (e.g. `CSUB`, `SBB`, jump tables, etc) we will need more space.  Some options:

* (**Folding Wide Opcodes**) One obvious is to fold the `WIDE` bit into the instruction opcodes themselves.  This will free up more encodings as not every opcode has a `WIDE` format.  However, many do, and this is perhaps not enough.  
* (**Variable-Length Opcodes**) Another option is to have a dedicated `WIDE` opcode which signals a `u16` opcode rather than a `u8` opcode.  This is a common trick used in other situations, and increases the space for encodings quite considerably.

Another observation here is that it would be useful to pull out constants into some kind of "constant pool" (at least, for the wide instructions).

DavePearce:
This PR tweakings the interpreter encodings to free up more opcode space.  This is done by introducing `u16` "wide opcodes" which are triggered by a small opcode `0xff`.  Thus, short-form encodings are unaffected whilst wide encodings have a minor performance impact.

### DIFF
--- a/pkg/zkc/vm/debug.go
+++ b/pkg/zkc/vm/debug.go
@@ -37,7 +37,9 @@ func BootAndDebug[W Word[W]](program Program[W], in map[string][]byte, observer 
 		fid, pc, frame := interpreter.ExtractExecutingState(bci)
 		// Clone the frame so the observer sees a stable snapshot which cannot be
 		// mutated by subsequent execution.
-		observer(State[W]{fid, pc, opcode == encoding.RET, slices.Clone(frame)})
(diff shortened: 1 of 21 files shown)


letypequividelespoubelles (approved):
blind approval

PR #2092, opened 2026-08-02
feat: tidy up keccak benchmark

### LINKED ISSUE #2091: fix: sort out Keccak benchmark
# Summary

Currently, we have multiple implementations of the keccak benchmark --- when we should only have one.  I think it makes sense to have a single benchmark which accepts a proper byte message and produces it hash, such that it can be compared with online calculators.

DavePearce:
This tidies up the keccak benchmark by removing the various variations that existed before, and settles on a single keccak benchmark which reads a byte message and computes the hash.  The input/output pairings can be calculated easily enough using an online calculator, or some external keccak implementation.

This also updates the implementation to match exactly that used in the current interpreter.  This just helps to ensure consistency between them.


OlivierBBB (approved):
LGTM

PR #2095, opened 2026-08-03
perf: don't factor skipIf when not required

### LINKED ISSUE #2093: fix: poor translation of `switch` on tracing path
# Summary

For reasons unknown, the translation of `switch` statements in the tracing path is introducing unnecessary binary lines and it would be good to prevent this.

## Example

Example to reproduce (adapted from `switch_09.zkc`:

```
input inputs(c:u1) -> (in:u32)

fn main() {
    var discr:u32 = inputs[0]

    switch discr {
        case 0xfa1709: {
            fail
        }
    }
}
```

Running this command:

```
go run ./cmd/zkc/main.go compile test.zkc
```

Produces the following:

```
input inputs(c:u1) -> (in'1:u16,in'0:u16)

fn main() -> ()
         u16 discr'1
         u16 discr'0
         u1 $1
         u1 $b_switch_1_case_1
         u1 $3
         u1 $b_switch_1_case_default
         u1 $5
 [00]  ldc $1 = 0x0
       read discr'1,discr'0 = inputs[$1]
       skip_if discr'1;discr'0 == fa;1709 2                      ─┐
       ldc $5 = 0x0                                               │
       skip 1                                                     │ ─┐
       ldc $5 = 0x1                                              ◄┘  │
       skip_if $5 != 0 2                                         ─┐ ◄┘
       ldc $b_switch_1_case_1 = 0x0                               │
       skip 1                                                     │ ─┐
       ldc $b_switch_1_case_1 = 0x1                              ◄┘  │
       ldc $3 = 0x1                                                 ◄┘
       sub $b_switch_1_case_default = $3-$b_switch_1_case_1
       check $b_switch_1_case_default:u1
       dispatch [$b_switch_1_case_1:1] $b_switch_1_case_default
       ret
       fail ""
```

What we can see is that we have introduced **two** binary lines to handle `case 0xfa1709` when only one is needed.  Specifically, we have `$5` and `$b_switch_1_case_1` where `$5 == $b_switch_1_case_1` always holds.

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/factor_skip_conditions.go
```diff
@@ -125,6 +135,46 @@ func generatesInverse[W word.Word[W]](si *bytecode.SkipIf[W], registers split.Al
...
 }
 
+// bodyContainsOnlyBitEquality reports whether the SkipIf at index i heads a diamond
+// which merely selects between two constants for a single 1-bit register:
+//
+//	skip_if (cond) 2
+//	b = k0
+//	skip 1
+//	b = k1
+func bodyContainsOnlyBitEquality[W word.Word[W]](codes []Bytecode[W], i uint, registers split.Allocator[W]) bool {
```
DavePearce, line 145: I agree this is not very pretty.  I do wonder if there is a better way of doing this.

DavePearce: Seems like this is currently matching the following lines from `process_B_type_instruction()`:

```
            match_case = registers[rs1] == registers[rs2] ? 1:0
```

DavePearce: But, I understand there is no point in factoring something which is already effectively factored.  So, I don't have a better suggestion at this time.

DavePearce (approved):
LGTM --- the stats are greatly improved by this.  I did leave a comment about `bodyContainsOnlyBitEquality` as, like you said, this feels a bit hacky --- but I understand why it is included.

pkg/zkc/vm/internal/transform/factor_skip_conditions.go
```diff
@@ -94,10 +94,20 @@ func factorableSkips[W word.Word[W]](codes []Bytecode[W], registers split.Alloca
 			factor[uint(i)] = false
 			continue
 		}
+		// Nothing to factorize if the body of the skip is a bit equality like b = x == 0 ? 1 :0.
+		// Note that as we lowerSwitch later, this pattern can't arise from lowerSwitch, but only
+		// directly from .zkc program.
```
OlivierBBB, line 99: Claude tells me that both `factorSkipIf` and `FactorLimbEqualities` emit that pattern

pkg/zkc/vm/internal/transform/factor_skip_conditions.go
letypequividelespoubelles: yeah, I was struggling a little bit as well to make it looks better (or at least less ugly)

pkg/zkc/vm/internal/transform/factor_skip_conditions.go
letypequividelespoubelles: and mainly lowerSwitch (~100 occurences for the r5 interpretor)

PR #2097, opened 2026-08-03
fix: don't pack shifted cells together

### LINKED ISSUE #2094: fix: incorrect constraint generation `bytecode.Dispatch`
# Summary

There is some unexplained problem with `bytecode.Dispatch` which prevents a constraint failure in certain situations.

## Example

Minimal example to reproduce (adapted from `switch_09.zkc`):

```
input inputs(c:u1) -> (in:u32)

fn main() {

    var discr:u32 = inputs[0]
    var res:u32 = 23

    switch discr {
        case 0xfa1709: {
            fail
        }
        case 0xbeef: {
            res = 0xffffffff
        }
        default: {
            res = 0x11111111
        }
    }

    if res != 0xffffffff && res != 0x11111111 {
        fail
    }
}
```
Input:
```
{ "inputs": "0x00fa1709" }
```

Executing this with the following command highlights the issue:

```
% go run ./cmd/zkc/main.go trace -cvv switch_09.json switch_09.zkc

DEBU[0000] including source file switch_09.zkc
DEBU[0000] Machine execution (19 steps) took 0.00s using 0 Gb (0 GC events) [0 Gb]
DEBU[0000] Trace processing took 0.00s using 0 Gb (0 GC events) [0 Gb]
DEBU[0000] Constraint compilation took 0.00s using 0 Gb (0 GC events) [0 Gb]
DEBU[0000] Expansion batch 0 (remaining 0) took 0.00s using 0 Gb (0 GC events) [0 Gb]
DEBU[0000] Trace expansion took 0.00s using 0 Gb (0 GC events) [0 Gb]
DEBU[0000] Trace generation took 0.01s using 0 Gb (0 GC events) [0 Gb]
DEBU[0000] Constraint batch 0 (110 items) took 0.02s using 0 Gb (3 GC events) [0 Gb]
DEBU[0000] Constraint checking took 0.02s using 0 Gb (3 GC events) [0 Gb]
ERRO[0000] machine panic
exit status 4
```

The key here is that the **constraints passed without generating a constraints failure**, even though the execution led to executing the `fail` statement.

**NOTE**: this requires #2038 to have been merged in order to see it correctly.

letypequividelespoubelles:


pkg/ir/mir/lower.go
```diff
@@ -493,19 +493,55 @@ func (p *AirLowering[F]) lowerConjunct(e *Conjunct[F], airMod air.ModuleBuilder[
...
-		// determine length of next conjunct
-		n := p.nextSumConjunct(airMod, worklist)
-		// construct next sum
-		sums = append(sums, term.Sum(worklist[:n]...))
-		// Remove n terms from worklist
-		worklist = worklist[n:]
+	// Sum only terms sharing identical evaluation bounds.
+	// If not, we may end up with constraint targeting oob cells,
+	// and the whole constraint would be dropped, instead of just the oob terms.
+	for _, group := range groupTermsByBounds(worklist) {
```
OlivierBBB, line 408:
```
	// If not, we may end up with constraint targeting oob cells,
	// and the whole constraint would be dropped,
```
Is this behaviour we are stuck with ?

OlivierBBB (approved):
LGTM

pkg/ir/mir/lower.go
OlivierBBB: Also: does zkc notify us if `the whole constraint would be dropped` ?

pkg/ir/mir/lower.go
```diff
@@ -493,19 +493,55 @@ func (p *AirLowering[F]) lowerConjunct(e *Conjunct[F], airMod air.ModuleBuilder[
...
 	}
-	//
-	for len(worklist) > 0 {
-		// determine length of next conjunct
-		n := p.nextSumConjunct(airMod, worklist)
-		// construct next sum
-		sums = append(sums, term.Sum(worklist[:n]...))
-		// Remove n terms from worklist
-		worklist = worklist[n:]
+	// Sum only terms sharing identical evaluation bounds.
```
DavePearce, line 405: So, this is related to clipping I guess ?

pkg/zkc/constraints/translator.go
```diff
@@ -394,7 +394,13 @@ func translateFunction[W vm.Word[W], F field.Element[F]](ctx schema.ModuleId, fn
 	// Translate all bytecode vectors
 	for pc, vec := range fn.Vectors() {
 		var (
-			handle = fmt.Sprintf("pc%d", pc)
+			handle = func() string {
+				if fn.IsOneLine() {
+					return "inst"
```
DavePearce, line 378: Hmmmm, interesting name `inst`.  I feel like `pc0` is better and more consistent, but its no big drama.

DavePearce (commented):
Overall, I am happy to approve this ... but I would like to talk it through in person first to be sure I understood it correctly.

pkg/ir/mir/lower.go
letypequividelespoubelles: >Is this behaviour we are stuck with ?

yes

> Also: does zkc notify us if the whole constraint would be dropped ?

not that I'm aware

pkg/ir/mir/lower.go
letypequividelespoubelles: yes

pkg/zkc/constraints/translator.go
letypequividelespoubelles: I can rollback, I just wanted to put pc+1 for MLI (as ivector kth is gated by IS_PC_(k+1), and after updated it for OLI. But ok to put back pc0 for OLI, I like both

PR #2098, opened 2026-08-03
test: add .rejects

letypequividelespoubelles:


DavePearce (approved):
This looks fine, though I wouldn't recommend trying to add any more `rejects` tests as their ability to find issues is quite limited still.

PR #2101, opened 2026-08-04
chore: tidy and remove pkgasm

### LINKED ISSUE #2055: chore: tidy and remove `pkg/asm`
# Summary

The package `pkg/asm` is no longer really used, and should be tidied and removed.  This will require moving some pieces over to the `pkg/zkc` package somewhere, to handle the few pieces which are now used by ZkC.  Furthermore, the concept of a `BranchId` can be generalised and simplified (e.g. to reduce the distinction between big endian and little endian forms).

DavePearce:


letypequividelespoubelles (approved):
Bye bye lisp

Issue #2069, opened 2026-07-28
feat: variable timestamp width for read-write-memory


OlivierBBB:
Make the timestamp variable width. Also use standard `@1, @2, ..., @n` bit columns + one `TIMESTAMP_DELTA` column to prove timestamp monotony.

DavePearce:
@OlivierBBB Please see suggested syntax from #1807, for example:

```
memory data(u16; addr:u16) -> (byte:u8)
```

Here, the first `u16` determines the width of the timestamp.  Alternative syntax could be e.g:

```
memory data[u16](addr:u16) -> (byte:u8)
```

Here, `[u16]` determines the width of the timestamp.  Every `memory` declaration should be required to provide a timestamp width.

PR #2104, opened 2026-08-04
feat: read-write memory (RAM) table constraints and trace observer (2/3)

DavePearce:
Part 2 of 3 of the RAM timestamps feature (#1807), stacked on #2079 (the diff shown here is only this PR's changes). Part 3 — the caller→memory lookups consuming the threaded stamps — follows.

## What it does

Gives every read-write memory a real constraint table and the trace observer which fills it: one row per access, in access order, carrying the address, the value the cell holds after the access (`VALUE_WRITTEN`), the value and timestamp it held before it (`VALUE_READ` / `TIMESTAMP_READ`, reconstructed by replaying the access log forward), the access's threaded timestamp (`TIMESTAMP_WRITTEN`) and a read/write flag. The `ramLayout` type in `constraints/ram.go` documents the role of every column — it is the specification of how the table works (including what the deltas are for: `TIMESTAMP_DELTA` witnesses the strict ordering `TS_READ < TS_WRITTEN`; `ADDRESS_DELTA` witnesses the strict address ordering of the future finalization phase, which is how each touched cell will be proven initialised/finalised exactly once).

Constraints enforced now: phase bits binary/exclusive/monotone; `TIMESTAMP_WRITTEN = TIMESTAMP_READ + 1 + TIMESTAMP_DELTA` as a multi-limb add-with-carry (timestamp columns split by the field's register width, exactly like the caller's threaded stamp registers — this is why the observer takes the field configuration); reads return what they find (`VALUE_READ == VALUE_WRITTEN`); range proofs on all columns. Deferred with the offline memory-checking bus (#2068): the receive/send argument that `VALUE_READ` really is the last value written, and the finalization rows.

## Review feedback addressed (from #2030)

- every column documented where it is declared (`ramLayout`), including the address/timestamp deltas;
- `Accesses()` renamed to `AccessLog()`;
- the observer's `field.Config` dependency documented: a 32-bit timestamp becomes e.g. two u16 columns on a 16-bit field, and must split exactly like the caller's stamp register — the underlying system handles arbitrary widths even though ZkC source cannot yet express them (#2069 tracks the source-level syntax);
- no bespoke test class: the `ram_01..10` fixtures run through the standard end-to-end framework (execution, gogen and constraint paths), now checked against the live RAM table — as are the RAM-heavy benches (blake / sort / keccakf / poseidon).

## Testing

New fixtures: `ram_09` (field-element values → native table columns) and `ram_10` (tuple values → one logical access spanning several lanes, regrouped by the observer). Full `Test_ZkcUnit|Test_ZkcMixed|Test_ZkcInvalid` and the full bench suite pass with the RAM table live; lint clean.


OlivierBBB (approved):
I ran constraints generation on both this and the original PR https://github.com/LFDT-Lineth/zkc/pull/2080: the constraints are identical with one exception in `ram_11` which seems to be related to switch refactorings.

LGTM

PR #2105, opened 2026-08-04
perf: REM/DIV with constant + `/%` new operation

### LINKED ISSUE #1863: perf: optimize register size when / and % by constant

### LINKED ISSUE #1876: perf: introduce /% zkc operator
When there is division and a remainder:
- within the same vector
- with the same registers (or constants) input

The hint for a division already produces the quotient and the remainder. No need to compute it twice for the two operations.

Then instead of creating two divison / remainder lowering pass, we should do only one:

```

fn div_and_remain(a:u8, b:u8) -> (q:u8, r:u8) {
q = a / b
r = a % b
}
```

is now compiling to:

```
fn div_and_remain(u8 a, u8 b) -> (u8 q, u8 r) {
        u8 $4
        u8 $5
        u8 $6
        u0 $7
        u0 $8
        u8 $9
        u8 $10
        u8 $11
        u0 $12
        u0 $13
[0]     q, $4, $5 = hint(a, b) ; $6 = q * b * 0x1 ; $7 = a - $6 - $4 ; $8 = b - $4 - $5 - 0x1 ; $9, r, $10 = hint(a, b) ; $11 = $9 * b * 0x1 ; $12 = a - $11 - r ; $13 = b - r - $10 - 0x1 ; ret
}
```

but should be only:

```
fn div_and_remain(u8 a, u8 b) -> (u8 q, u8 r) {
        u8 $5
        u8 $6
        u0 $7
        u0 $8
[0]     q, r, $5 = hint(a, b) ; $6 = q * b * 0x1 ; $7 = a - $6 - $4 ; $8 = b - $4 - $5 - 0x1 ; ret
}
```

letypequividelespoubelles:


pkg/zkc/compiler/codegen/statement.go
```diff
@@ -733,45 +733,84 @@ func (p *StmtCompiler) compileIntDiv(args []Expr, bitwidth uint, mapping []uint,
...
+	var (
+		value = sources[0]
+		last  = len(sources) - 1
+	)
 	//
-	for i := 1; i < len(sources)-1; i++ {
-		tmp := p.allocate(util.Some(bitwidth))
-		insns = append(insns, vm.Div[vm.Uint](tmp, value, sources[i]))
-		value = tmp
+	for i := 1; i <= last; i++ {
```
DavePearce: Tbh, this is getting pretty icky.  Whilst I am satisfied that the translation here looks good (i.e. with appropriate overflow checks) I personally think its just better to avoid this situation altogether.  That is, make it impossible to write `a / b / c` and force the user to write something like `(a / b) / c` instead.  That way, you can just ignore folding here altogether because, in practical reality, I have *never* seen anyone write a multi-operand division.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -733,45 +733,84 @@ func (p *StmtCompiler) compileIntDiv(args []Expr, bitwidth uint, mapping []uint,
...
-	sources, insns := p.compileUniformArgs(bw, mapping, args...)
-	// Chain remainders left-to-right: (((a % b) % c) % ...).
-	value := sources[0]
-	//
-	for i := 1; i < len(sources)-1; i++ {
-		tmp := p.allocate(bw)
-		insns = append(insns, vm.Rem[vm.Uint](tmp, value, sources[i]))
-		value = tmp
+	// Compile the dividend upfront.
+	value, insns := p.compileArg(args[0], bw, mapping)
```
DavePearce, line 736: Ditto regarding the need to support `a % b % c`.

pkg/zkc/vm/bytecode.go
```diff
@@ -484,12 +485,24 @@ func Div[W Word[W]](target, dividend, divisor RegisterId) Bytecode[W] {
...
+
 // Rem constructs an integer-remainder instruction computing
 // "target = dividend % divisor".
 func Rem[W Word[W]](target, dividend, divisor RegisterId) Bytecode[W] {
 	return bytecode.NewDivRem[W](encoding.REM, target, dividend, divisor)
 }
 
+// RemConst constructs an integer-remainder instruction computing
+// "target = dividend % divisor" for a constant divisor.
+func RemConst[W Word[W]](target, dividend RegisterId, divisor W) Bytecode[W] {
```
DavePearce: You could consider having a single `Rem()` function which accepts an `Operand[W]`, like for `SkipIf`.

pkg/zkc/vm/internal/gogen/emit_bitwise.go
```diff
@@ -155,50 +154,6 @@ func (g *generator) pairCall(c *code, helper string, lhs, rhs operand, target li
 	return inner
 }
 
-// emitDivRem emits DIV / REM (executeDiv/Rem): a zero divisor fails, otherwise
```
DavePearce, line 158: Looks like an improvement to put it in the right place.

pkg/zkc/vm/internal/interpreter/encoding/common.go
```diff
@@ -241,6 +241,10 @@ const (
 	DEBUG
 	// FIELD_TO_UINT instruction
 	FIELD_TO_UINT
+	// DIVC (divide by constant) instruction
```
DavePearce: Maybe instead of sticking them at the end, put them with the other similar ones (e.g. `ADDC`, `MULC`, etc)

pkg/zkc/vm/internal/interpreter/encoding/div.go
```diff
@@ -18,28 +18,58 @@ import (
...
-	return encodeIntrinsic(p.Op, p.Targets, p.Sources)
+func Intrinsic[W word.Word[W]](p *bytecode.Intrinsic[W], env Environment[W]) []uint32 {
+	return encodeIntrinsic(p.Op, p.Targets, p.Sources, env)
 }
 
-// DecodeIntrinsic decodes an intrinsic instruction at the given program counter.
-func DecodeIntrinsic[W word.Word[W]](pc uint32, codes []uint32) (Bytecode[W], uint32) {
+// DecodeIntrinsic decodes an intrinsic instruction at the given program
+// counter, resolving constant sources against the given constant pool.
+func DecodeIntrinsic[W word.Word[W]](pc uint32, codes []uint32, pool []W) (Bytecode[W], uint32) {
```
DavePearce: This function is not actually used, and can be removed I think.

pkg/zkc/vm/internal/interpreter/encoding/div.go
```diff
@@ -18,28 +18,58 @@ import (
...
 	)
 	//
 	return &bytecode.Intrinsic[W]{Op: op, Targets: targets, Sources: sources}, n
 }
 
+// operandsFromIter reconstructs the source operands packed as flagged (base,
+// len) pairs within the given iterator (see Operands.NextOperand), resolving
+// constants against the given constant pool.  Constant operands are
+// single-limb by construction (enforced when encoding, see encodeIntrinsic).
+func operandsFromIter[W word.Word[W]](iter Operands, pool []W) []bytecode.Operand[W] {
```
DavePearce: Ditto --- can be removed.

pkg/zkc/vm/internal/interpreter/encoding/div.go
```diff
@@ -107,6 +137,48 @@ func DecodeDivRem_2n1(pc uint32, codes []uint32) (rd, dividend, divisor Register
...
 }
 
+// ============================================================================
+// DIVC / REMC (constant divisor) instruction.  The format mirrors the
+// arithmetic-with-constant family exactly (see encodeArith_1n1c), with rs
+// holding the dividend and imm8 the small constant divisor:
+//
+//	31                                0
+//
+// +--------+--------+--------+--------+
```
DavePearce, line 106: Looks good.

pkg/zkc/vm/internal/bytecode/div.go
```diff
@@ -18,21 +18,28 @@ import (
...
 	// Opcode selects the operation (DIV or REM).
 	Opcode uint32
 	// Target receives the result.
 	Target RegisterId
-	// Dividend and Divisor are the operand registers.
-	Dividend, Divisor RegisterId
+	// Dividend is the operand register.
+	Dividend RegisterId
+	// Divisor is either a (single limb) operand register or a constant.
+	Divisor Operand[W]
```
DavePearce, line 33: It is a bit odd for me that you only change the divisor to be an `Operand[W]`, but can do that later if/when the need arises.

pkg/zkc/vm/internal/interpreter/encoding/common.go
```diff
@@ -365,6 +369,11 @@ const (
 	WIDE_DEBUG
 	// WIDE_FIELD_TO_UINT instruction
 	WIDE_FIELD_TO_UINT
+	// WIDE_DIVC (divide by pooled constant) instruction
+	WIDE_DIVC
```
DavePearce: Ditto

pkg/zkc/vm/internal/interpreter/interpreter.go
```diff
@@ -1491,16 +1529,23 @@ func shr64[W word.Word[W]](values []W, width uint64) []W {
...
+		base, length, isConst = iter.NextOperand()
+		zero                  W
+		bandwidth             = zero.Bandwidth()
+		words                 []W
+		cur                   W
+		off                   uint
 	)
+	// A constant operand is single value by construction.
+	if isConst {
+		if length != 1 {
```
DavePearce: To me, it seems odd that we only support a single limb here.  I don't get that.

DavePearce: Yeah, the restriction in `split.Operand()` seems odd.

pkg/zkc/vm/internal/transform/split/divrem.go
```diff
@@ -51,16 +56,91 @@ func DivRem[W word.Word[W]](mapping descriptor.LimbsMap[W], insn *bytecode.DivRe
...
+		[]bytecode.Operand[W]{
+			bytecode.NewRegisterVectorOperand[W](dividend),
+			divisor,
+		}))
+}
+
+// Operand splits an operand into limbs: a register operand becomes the
+// limb vector of its constituent registers, whilst a constant operand stays a
+// single (unsplit) constant when it fits the register width — the invariant
+// being that constant operands are always single-limb.  A constant too wide
```
DavePearce: These requirements seem wierd to me.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -124,32 +163,275 @@ func expandRemainder[W word.Word[W]](r, x, y bytecode.RegisterId, registers spli
...
 		// (q = x/y, r = x%y), equals the dividend x exactly and so fits nX
 		// bits; the extra bit guards the addition's transient carry.
-		qyr = registers.Allocate("", util.Some(nX+1))
-		rw1 = registers.Allocate("", util.Some(nY+1))
-		// NOTE: must separate z0 & z1 to avoid write conflict (for now).
-		z0 = registers.Allocate("", util.Some[uint](0))
-		z1 = registers.Allocate("", util.Some[uint](0))
+		qyr = registers.Allocate("qy_plus_r", util.Some(nX+1))
+		// rw1 = r + w + 1 holds the divisor exactly (nY bits) + 1 to ensure no overflow
+		nRW = max(registers.Register(r).Bitwidth().Unwrap(),
```
DavePearce: This is an interesting choice to change, and I would leave as `util.Some(nY+1)`.  There is no need to guard against overflow as, when this happens, we want a constraint failure.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -124,32 +163,275 @@ func expandRemainder[W word.Word[W]](r, x, y bytecode.RegisterId, registers spli
...
+
+		mulQY, subZ1 Bytecode[W]
 	)
 	//
+	if y.IsConstant() {
+		mulQY = bytecode.MulConst(qy, []bytecode.RegisterId{q}, y.AsConstant())
+		subZ1 = bytecode.SubConst(z1, []bytecode.RegisterId{rw1}, y.AsConstant())
+	} else {
+		mulQY = bytecode.MulConst(qy, []bytecode.RegisterId{q, y.AsRegister()}, one)
+		subZ1 = bytecode.SubConst(z1, []bytecode.RegisterId{y.AsRegister(), rw1}, zero)
```
DavePearce, line 139: I think some kind of special instruction to manage this form at some point might be useful.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -124,32 +163,275 @@ func expandRemainder[W word.Word[W]](r, x, y bytecode.RegisterId, registers spli
...
+		subZ1,
 	}
 }
+
+// ============================================================================
+// DIV / REM pair merging
+// ============================================================================
+
+// bcPosition identifies a bytecode by its vector index and its index within
+// that vector.
```
DavePearce: I have to say that I am a little unsure why you are going to so much effort for this ... is it really giving us a big pay off?

DavePearce: To be honest, I would actually approach this differently.  For the record, this is what I would do: (1) split `Target` in `bytecode.DivRem` into a `Quotient` and `Remainder` components, both of which are `util.Option` (where we always require one, and sometimes have two); (2) we could add a special statement form, such as `q,r = a / b` and/or move the planning logic here into a separate transform whose sole purpose is to do the matching, and nothing else.

DavePearce: The point being to spread out the complexity.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -124,32 +163,275 @@ func expandRemainder[W word.Word[W]](r, x, y bytecode.RegisterId, registers spli
...
+//     vector which is the target of a jump);
+//   - neither operand is redefined; and
+//   - the second member's target is neither read nor written (its write moves
+//     to the first member's position).
+//
+// The scan is a single linear pass maintaining the unpaired DivRems seen so
+// far, conservatively cleared at every control-flow bytecode and jump target.
+// Same-opcode duplicates are not merged (that requires a copy bytecode); the
+// first occurrence stays the pairing candidate.
+func planDivRemMerges[W word.Word[W]](fn *descriptor.Function[W]) divRemPlan[W] {
```
DavePearce: I am not a fan of this, sorry.  This is a lot of stuff for something we may not need.  If we absolutely do need it, then easiest easiest way around this would be to add a dedicated instruction which allows us to extract the quotient and remainder at the same time.  That way we can avoid this kind of ad-hoc static analysis.  Claude has already proved itself to be bad at this kind of work.

pkg/zkc/vm/internal/transform/optimize_division.go
```diff
@@ -1,214 +0,0 @@
-// Copyright Consensys Software Inc.
```
DavePearce, line 1: Its good to see we are at least losing stuff.

pkg/zkc/vm/internal/transform/lower_division.go
DavePearce: See this: https://docs.python.org/3/library/functions.html#divmod

pkg/zkc/vm/internal/transform/split_registers.go
```diff
@@ -525,3 +529,26 @@ func splitRegisterVectors[W any](limbsMap descriptor.LimbsMap[W],
 	//
 	return nvecs
 }
+
+// splitOperandVectors splits each operand (e.g. an intrinsic argument) into
```
DavePearce, line 531: This is so wierd.  I guess this is something to do with the fact that a constant operand does have bitwidths for the limbs??  But, in fact, we know what their bitwidths are.

pkg/zkc/vm/internal/transform/lower_division.go
DavePearce: We could use a syntax like `q,r = x /% y`.  Or even abuse the syntax from python: `q,r = x // y` --- which is a bit neater.

DavePearce (commented):
I think we talk at the daily to go through this.  Its a bit difficult to understand all that is going on here.  It seems actually like two PRs: (1) support constant divisors on `DivRem`; (2) combine `q = x/y; r = x%y` into one `q,r = x%y` to eliminate duplication.  For (1), this actually seems fine to me (modulo the issue around supporting only a single limb); For (2), it seems like a lot of effort for something which is only used once in the arithmetisation (technically, its used twice --- but the second use can be rewritten).  The simpler solution to support (2) is to introduce a dedicated expression form, such as `x//y` as I suggested in the comments.  Anyway, this could still be merged as is, and then we put in the `//` expression separately later so we can remove `planDivRemMerges()`.

pkg/zkc/vm/internal/gogen/emit_bitwise.go
letypequividelespoubelles: yes ..., just cleaning but yeah

pkg/zkc/compiler/codegen/statement.go
letypequividelespoubelles: do agree ...

pkg/zkc/compiler/codegen/statement.go
letypequividelespoubelles: do agree ...

pkg/zkc/vm/internal/transform/lower_division.go
letypequividelespoubelles: thanks for pushing me do it, it's wayyyyyy cleaner

letypequividelespoubelles: // is for comment n zkc, don't want to change it

pkg/zkc/vm/internal/interpreter/interpreter.go
letypequividelespoubelles: > single limb here

It's not a single limb, it's a single constant (tajt ca be spli in multiple limbs)

pkg/zkc/vm/internal/bytecode/div.go
letypequividelespoubelles: hum ... you mean to accept cst / b ?

pkg/zkc/vm/internal/transform/lower_division.go
letypequividelespoubelles: I don't understand what you mean here ?

pkg/zkc/vm/internal/transform/check_casts.go
```diff
@@ -91,7 +91,7 @@ func castPacket[W word.Word[W]](b Bytecode[W], regmap descriptor.RegisterMap[W],
 			width = dividend.Bitwidth()
 		}
 		//
-		return prepend(b, checkCast(regmap, width, b.Target))
+		return prepend(b, checkCast(regmap, width, b.Definitions()...))
 	case *bytecode.Call[W]:
```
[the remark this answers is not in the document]
letypequividelespoubelles: REM_DIV has been lowered already in tracing mode, so I guess we don't care

pkg/zkc/compiler/ast/expr/divmod.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/LFDT-Lineth/zkc/pkg/zkc/compiler/ast/variable"
+)
+
+// DivMod represents the combined division/remainder ("/%") expression, which
+// computes both the quotient and the remainder of its two operands.  Unlike
+// the other arithmetic operators it produces two values, so its type is a
+// tuple and it can only appear as the full source of a two-target assignment
+// (e.g. "q, r = a /% b").
+type DivMod[S symbol.Symbol[S]] struct {
+	Exprs    []Expr[S]
```
DavePearce, line 29: Seems to me it would be better not to have an array here, but rather make the dividend and divisor explicit.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -703,75 +704,78 @@ func (p *StmtCompiler) compileFieldMul(args []Expr, mapping []uint, target Regis
...
-
-				break
-			}
-		} else {
-			nargs = append(nargs, e)
+	var bw = util.Some(bitwidth)
+	// Compile the dividend upfront.  Division is strictly binary (the parser
+	// rejects unbraced chains), so args holds exactly the dividend and divisor.
+	value, insns := p.compileArg(args[0], bw, mapping)
+	divisor, extra := p.compileOperand(args[1], bw, mapping)
```
DavePearce, line 711: That is nicer.

pkg/zkc/compiler/codegen/statement.go
```diff
@@ -703,75 +704,78 @@ func (p *StmtCompiler) compileFieldMul(args []Expr, mapping []uint, target Regis
...
+		} else if divisor.AsConstant().Cmp64(1) == 0 {
+			p.errors = append(p.errors, p.srcmaps.SyntaxErrors(e.Exprs[0], "division has no divisor")...)
+		}
+	}
 	//
-	for i := 1; i < len(sources)-1; i++ {
-		tmp := p.allocate(bw)
-		insns = append(insns, vm.Rem[vm.Uint](tmp, value, sources[i]))
-		value = tmp
+	if p.fastMode {
```
DavePearce: Ok, this will be a problem at some point.  Personally, I don't see why its even necessary either?  But, regardless, the issue is that we need `codegen` to build a neutral `vm` program which will be lowered for its intended use case **after** `codegen` has completed.  One way to think about this is that we will be moving the lowering pipeline in `compile.go` (i.e. which distinguishes fast mode from tracing mode) out of `codegen` altogether.  This is to facilitate a single program which can be used for either tracing or fast mode (i.e. because tracing actually requires both).  Anyway, I can fix this when I make that change --- its just an FYI for now.

pkg/zkc/compiler/validate/typing.go
```diff
@@ -254,6 +254,10 @@ func (p *TypeChecker) typeAssignment(s *stmt.Assign[symbol.Resolved], env Variab
 		//
 		return errors
 	}
+	// A divmod produces exactly the (quotient, remainder) pair.
+	if _, ok := s.Source.(*expr.DivMod[symbol.Resolved]); ok && len(s.Targets) != 2 {
```
DavePearce: This check seems out-of-place, and I don't like special cases like this.  I think it could be moved into `typeDivMod` without any problem.  Also, another option would be to update `checkEquiTypes()` to report a type error as follows:
```
return p.srcmaps.SyntaxErrors(node, fmt.Sprintf("expected type %s, found %s", rhs.String(p.env), lhs.String(p.env)))
```
That improves the error messages in general for DivMod, but also for other situations like function calls.

pkg/zkc/vm/internal/bytecode/div.go
```diff
@@ -15,47 +15,86 @@ package bytecode
...
+// divisor aborts execution with a division-by-zero error.
 type DivRem[W word.Word[W]] struct {
-	// Opcode selects the operation (DIV or REM).
-	Opcode uint32
-	// Target receives the result.
-	Target RegisterId
-	// Dividend and Divisor are the operand registers.
-	Dividend, Divisor RegisterId
+	// Quotient receives the division result, when present.
+	Quotient util.Option[RegisterId]
```
DavePearce: So, this is what I was suggesting before, except --- they don't even need to be options, because: (1) on the tracing side they must exist anyway; (2)  on the fast mode side, well we don't really care if we over allocate by one register (i.e. as `gogen` will get rid of it anyway when running the `go` compiler).  So, avoiding having them as `util.Option` just simplifies a lot the code which uses this type.

pkg/zkc/vm/internal/bytecode/div.go
DavePearce: Yeah, for example.  But, there are no active use cases for this.  I'm just imagining some might come in the future.  And, we can fix is/when they do arise.

pkg/zkc/vm/internal/bytecode/div.go
```diff
@@ -15,47 +15,86 @@ package bytecode
...
 		dividend = RegisterToString(p.Dividend, mapping)
-		divisor  = RegisterToString(p.Divisor, mapping)
-		symbol   = "/"
+		divisor  = p.Divisor.String(mapping)
 	)
 	//
-	if p.Opcode != 0 {
-		symbol = "%"
+	switch {
+	case p.Quotient.HasValue() && p.Remainder.HasValue():
```
DavePearce: Lots of cases because of options.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -56,54 +55,61 @@ func lowerDivisionFunction[W word.Word[W]](fn *descriptor.Function[W]) *descript
...
+		w    = registers.Allocate("", util.Some(nY))
+		q, r bytecode.RegisterId
 	)
 	//
-	return expandDivRem(q, r, w, x, y, nX, nY, registers)
+	if !dr.Quotient.HasValue() && !dr.Remainder.HasValue() {
+		panic("At least one of quotient or remainder must be defined")
+	}
+	//
+	if dr.Quotient.HasValue() {
```
DavePearce: More complexity because of the choice to use `util.Option` for `Quotient` and `Remainder`.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -124,32 +130,49 @@ func expandRemainder[W word.Word[W]](r, x, y bytecode.RegisterId, registers spli
...
+		qyr  = registers.Allocate("qy_plus_r", util.Some(nX))
+		// rw1 holds r + w + 1, which the zero-assert below forces to equal the
+		// divisor y — together with w's range constraint this establishes
+		// r < y.  It is sized to the widest addition operand: the sum then
+		// statically overshoots by at most one transient carry bit, which the
+		// splitter checks to zero, so any dishonest overflow is a constraint
+		// failure rather than a wrap.  With a freshly allocated remainder
+		// (lone division) this is the divisor width; with a user remainder
+		// target (remainder / divmod) it is the operand width nX.
+		rw1 = registers.Allocate("", util.Some(max(
```
DavePearce: It is interesting that we have so much written here (i.e. the comment), and yet it is still seems mistaken.  We know that, in the only case we care about, `r+w+1 == y` and, therefore, by construction it requires at most `nY` bits.  The transient carry argument (i.e. why we previously had `nY+1` as the width for `rw1`) may have held true at some point, but it was always a bug in the register splitting algorithm.  It should not be true now.  An example where this really is ugly would be:

```
fn f(x:u16, y:u16) -> (a:u32) {
   a = (x % y) as u32
}
```

Here, the `rw1` register is allocated as a `u32` --- when it does not need to be.

DavePearce: Here's a ZkC version of the assertion itself:

```
fn assert_r_lt_y(y:u16, r:u32, w:u16) {
   var rw1:u16 = (r as u16)+w+1
   //
   if y != rw1 {
      fail
   }
}
```

The question is: is there any situation under which this either: (1) fails when `r < y` holds; or (2) succeeds when `r >= y`.

DavePearce: If the answer is "yes" to either of these then we have a problem which needs to be fixed somewhere.

DavePearce: Also, since you are using the prefix functionality, you might as well give `rw1` its `"rw1"` prefix.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -56,54 +55,61 @@ func lowerDivisionFunction[W word.Word[W]](fn *descriptor.Function[W]) *descript
...
 	//
-	return expandDivRem(q, r, w, x, y, nX, nY, registers)
+	if !dr.Quotient.HasValue() && !dr.Remainder.HasValue() {
+		panic("At least one of quotient or remainder must be defined")
+	}
+	//
+	if dr.Quotient.HasValue() {
+		q = dr.Quotient.Unwrap()
+	} else {
+		q = registers.Allocate("", util.Some(nX))
```
DavePearce: You might as well use a prefix here like e.g. `"q"`.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -56,54 +55,61 @@ func lowerDivisionFunction[W word.Word[W]](fn *descriptor.Function[W]) *descript
...
 	var (
-		nX = registers.Register(x).Bitwidth().Unwrap()
-		nY = registers.Register(y).Bitwidth().Unwrap()
-		r  = registers.Allocate("", util.Some(nY))
-		w  = registers.Allocate("", util.Some(nY))
+		x    = dr.Dividend
+		y    = dr.Divisor
+		nX   = registers.Register(x).Bitwidth().Unwrap()
+		nY   = divisorWidth(y, registers)
+		w    = registers.Allocate("", util.Some(nY))
```
DavePearce: You might as well use a prefix here like `"w"` since you have in other places.

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -56,54 +55,61 @@ func lowerDivisionFunction[W word.Word[W]](fn *descriptor.Function[W]) *descript
...
+	if dr.Quotient.HasValue() {
+		q = dr.Quotient.Unwrap()
+	} else {
+		q = registers.Allocate("", util.Some(nX))
+	}
+	//
+	if dr.Remainder.HasValue() {
+		r = dr.Remainder.Unwrap()
+	} else {
+		r = registers.Allocate("", util.Some(nY))
```
DavePearce: ditto re: prefix

pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -124,32 +130,49 @@ func expandRemainder[W word.Word[W]](r, x, y bytecode.RegisterId, registers spli
...
+		// statically overshoots by at most one transient carry bit, which the
+		// splitter checks to zero, so any dishonest overflow is a constraint
+		// failure rather than a wrap.  With a freshly allocated remainder
+		// (lone division) this is the divisor width; with a user remainder
+		// target (remainder / divmod) it is the operand width nX.
+		rw1 = registers.Allocate("", util.Some(max(
+			registers.Register(r).Bitwidth().Unwrap(),
+			registers.Register(w).Bitwidth().Unwrap())))
+		// TODO: must separate z0 & z1 to avoid write conflict (for now).
+		z0 = registers.Allocate("zero", util.Some[uint](0))
```
DavePearce: I actually think you can: (1) use `registers.ZeroRegister()` here; and (2) allocate only one zero register (i.e. `z` instead of `z0` and `z1`).  The reason for the `TODO` comment (which I wrote) is no longer valid since we do have a special case for a zero register, such that it never causes a write conflict.

pkg/zkc/vm/internal/transform/lower_division.go
DavePearce: I ask Claude ... let's see what it comes back with.

DavePearce: > The width is over-provisioned whenever width(r) > width(y)
> ...
> **Why max is presumably there anyway**
> 
> Sizing `rw1` to `nY` while `r` is wider gives the addition more source chunks than target limbs, so you'd land on the splitter's "target stack exhausted" path. `initialiseSubChunks` has an explicit overhang loop for surplus targets, but surplus sources with no target is the direction I'd want to check before trusting it. That's the real obstacle — not the arithmetic.

It is agreeing with the general hypothesis that `rw1` can be allocated `nY` bits, but is also suggesting a possible bug in register splitting for subtraction where it doesn't assign additional source registers as `0`.

DavePearce: It is interesting though that it then (for some reason) seems to ignore the bug and suggest a slightly more ridiculous option:

> If you want the win, the cleaner route is probably to narrow `r` first rather than widen `rw1`

I don't understand what its trying to say here.

DavePearce: > % Are you suggesting there is a bug in register splitting for subtraction when you said this: "but surplus sources with no target is the direction I'd want to check before trusting it." ?

> No — and having now read the code, my caveat was wrong on two counts. Thanks for pushing on it. ... But it's handled anyway. `RegisterStack.SelectExact` already covers the exhausted-stack case:
> ...
> So the conclusion I drew stands, but for a better reason than I gave: narrowing rw1 directly trades limbs for compounding carries.

So, it still believes there is a problem, but not for the reason indicated initially.  Now it thinks there is a problem with addition and "compounding carries".

DavePearce: > % That now sounds to me like you're suggesting a bug in the handling of addition which has something to do with these "compounding carries".  Can you clarify this?

> Fair challenge — I've speculated twice now. Let me test it instead.
> That's decisive — and it's a real, reachable panic in the current addition splitter, unrelated to my expandDivRem suggestion
> ... [some gibberish] ..
> Yes — but not the bug I was gesturing at, and it's in code you already ship, not in my proposal. Here's the confirmed version

So, seems like we have a bug.  I will raise a separate issue for this.  But, now after all this, look at its conclusion:

> This doesn't change the `rw1` analysis — `max(width(r), width(w))` keeps `expandDivRem` clear of the bad shape, which is part of why it's sized that way. But it does strengthen the recommendation: narrowing `rw1` directly would drive additions into exactly this path, so narrowing r instead is the safer route, and this panic wants fixing on its own merits first.

Basically, it wants to do something which is suboptimal to avoid a pre-existing bug, rather than just fixing the bug and doing the optimal choice.  This is driving me nuts.

DavePearce: > % To clarify, there is pre-existing and unrelated bug around addition (nice find) --- but your suggestion is to use a sub-optimal `bitwidth` for `rw1` to avoid this path?  That doesn't make sense to me, since its highly likely that sub-optimal choice would remain in the code base for some time to come.  Instead, let's assume we fixed the pre-existing bug in addition.  Would it then make sense to always allocate `rw1` with `nY` bits ?

> You're right, and I'll drop that recommendation — sizing a register sub-optimally to dodge a bug you're about to fix is exactly the kind of workaround that outlives its justification.
>
> Yes, `nY` is the correct width. `rw1` exists solely to be asserted equal to y on the next instruction. Its value is provably `y`, so `nY` bits is both sufficient and minimal:

Finally.  Let me now raise a separate issue for the bug.

DavePearce: https://github.com/LFDT-Lineth/zkc/issues/2117

DavePearce (changes requested):
Overall, looks good.  Have mostly very minor comments.  Only required change is for the bitwidth of `rw1` --- see detailed comments regarding "discussion" with Claude about this issue.

pkg/zkc/vm/internal/transform/lower_division.go
letypequividelespoubelles: indeed

OlivierBBB (approved):
Blind approval

DavePearce (approved):
Yup, LGTM

PR #2110, opened 2026-08-05
perf: O(log2(width)) instead of O(width) for shift operation

### LINKED ISSUE #1740: feat: optimise bitwise shift constraints in ZkC
## Summary

Currently bitwise shift operations are implemented using one row per shifted bit, which is relatively expensive (e.g. 63 rows for a 63 bit left shift).  On the positive side, the implementation is recursive meaning reuse between rows is possible (though whether or not this is likely to happen remains unclear).

## Example

The approach taken in `zkasm` was a recursive procedure (similar to the original `shf` module) where, for a shift `w << n`, each layer handles the shift for a specific bit of `n`.  For example, consider this function from the original `zkasm` implementation:

```
fn bit_shl256_u3(word u256, n u3) -> (res u256) {
  var lsw u252
  var msw u4
  var b u1
  var m u2
  ;; decompose shift
  b,m = n
  ;;
  if b!=0 goto apply
  res = bit_shl256_u2(word,m)
  return
apply:
  msw, lsw = word
  res = bit_shl256_u2(lsw * 2^4, m)
  return
}
```

This implements shifts on `u256` of upto `n=7` positions.  To understand this, consider the bit pattern for `n`.  If we have `0b1XX` then it performs `x << 4` to implement the "coarse" shift and then calls `bit_shl256_u2()` to get the "fine" shift.

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/lower_bitwise.go
```diff
@@ -142,104 +146,23 @@ func maxBitwidthOf[W word.Word[W]](regs []descriptor.Register[W], targets ...byt
...
+		return "not"
 	case bytecode.OP_SHL:
-		op = "shl"
+		return "shl"
 	case bytecode.OP_SHR:
-		op = "shr"
+		return "shr"
 	default:
-		op = "unknown"
+		return "unknown"
```
DavePearce: Should be a `panic` ?

pkg/zkc/vm/internal/transform/lower_shift.go
```diff
@@ -65,95 +75,267 @@ func scanShiftAmountWidths[W word.Word[W]](modules []descriptor.Module[W]) map[s
...
+// builder but NOT emitting (so the caller can size a Skip over the sequence):
 //
-//	shr(a, 0)    = a
-//	shr(a, n>=w) = 0
-//	shr(a, n)    = shr(floor(a/2), n-1)
+//	SHR: Destruct a into [drop:u_shift, high:u(width-shift)]; target = high
+//	     (zero-extended via AddConst, since high < 2^(width-shift)).
+//	SHL: Destruct a into [low:u(width-shift), drop:u_shift]; target = zeros:low
+//	     via Concat with a constant-zero register in the low bits.
+func shiftByConst[W word.Word[W]](b *helperBuilder[W], op bytecode.Operation,
```
DavePearce, line 261: At some point, we will support constants on bitwise bytecodes (and in other places, like function calls).  This would mean we could emit a real implementation for "shift by constant", instead of loading a constant into a register, and then doing a reg-reg shift etc.

DavePearce (dismissed):
LGTM --- very minor comments only.

pkg/zkc/vm/internal/transform/lower_shift.go
letypequividelespoubelles: yes

pkg/zkc/vm/internal/transform/lower_bitwise.go
letypequividelespoubelles: absolutely

DavePearce (approved):
LGTM

PR #2114, opened 2026-08-06
feat: simplify lookups

### LINKED ISSUE #2111: feat: simplify lookup constraints to support registers only
# Summary

The purpose of this feature is to effectively downgrade lookup constraints accepting vectors of arbitrary terms to just accepted vectors of registers.  This can be done because `zkc` does not require anything other than vectors of registers.  Furthermore, doing so simplifies the "signature" of a lookup which, in turn, will allow us (more easily) to optimise evaluation of lookup constraints.

DavePearce:
Lookup sources, targets and selectors must now be column accesses rather than arbitrary expressions, simplifying the parser, resolver and translator since a lookup no longer synthesises computed columns for its operands.  Every test which relied on an expression in a deflookup / defclookup has been rewritten to introduce a temporary register holding the value of that expression, defined by an ordinary vanishing constraint, with the corresponding accepts / rejects traces updated to supply the new column; constant selectors (as in valid/lookup_23 and agnostic/lookup_06) become columns with ":padding 1" so that the implicit padding row behaves exactly as before.  Each reject trace was checked to still fail because of its lookup, rather than because of the constraint defining its temporary register.


letypequividelespoubelles (approved):
almost blind approval

PR #2125, opened 2026-08-11
feat: support tracing for n occurences of a given function

### LINKED ISSUE #2102: feat: support tracing for `N` occurences of a given function
# Summary

The purpose of this feature is to enable a trace to be generated for only `N` occurrences of a given function.  This feature needs to be accessible through the API as well (e.g. so it can be accessed via the `zkcdriver`).

DavePearce:


OlivierBBB (approved):
Mostly blind approval

PR #2126, opened 2026-08-11
fix: bug with splitting addition / concat

### LINKED ISSUE #2117: fix: bug in register splitting for addition
(_see also #2056_)

# Summary

There is a bug in the register splitting algorithm for addition which can lead to a `panic` under the right circumstances.

## Example

A minimal example is:

```
fn f(x:u48) {
   var t:u16 = (x as u16) + 1
   return
}
```

Compiling this with the command:

```
% zkc compile broken.zkc
``` 

Gives the following error:

```
panic: register exceeds maximum width (17 > 16)

goroutine 1 [running]:
github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/transform/split.(*registerAllocator[...]).Allocate(0x106235de0, {0x105501d42, 0x1061377c0?}, {0x78?, 0x8f2953ccd80?})
	/Users/djp/projects/zkevm/go-corset-copy/pkg/zkc/vm/internal/transform/split/allocator.go:95 +0x228
github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/transform/split.insertAddCarryLines[...]({0x106214b18, 0x8f2953ccd80}, {0x8f294ce2500, 0x3, 0x4})
	/Users/djp/projects/zkevm/go-corset-copy/pkg/zkc/vm/internal/transform/split/addition.go:116 +0x160
github.com/LFDT-Lineth/zkc/pkg/zkc/vm/internal/transform/split.Addition[...]({0x12d289200, 0x8f2953d0460?}, {0x106214b18?, 0x8f2953ccd80}, 0x104d6d830)
	/Users/djp/projects/zkevm/go-corset-copy/pkg/zkc/vm/internal/transform/split/addition.go:53 +0x84
```

Another example is the following:

```
fn f(x:u64) {
   var t:u16 = x as u16
   return
}
```

## Analysis

With `KOALABEAR_16` the register width is `u16`, so `x:u48` splits into three limbs and the addition gets 3 chunks against a 1-limb target:

```
var c0:i1, c1:u17
var z:u0

c0::t = x'0 + 1
c1::z = x'1 + c0
z = x'2 + c1
```

The issue is that the size of the carry `c1` is calculated as `u17` when, in fact, we should have no carry at all for remaining chunks.

## Fixes

Proposed fix is to add a guard in `insertAddCarryLines()` like so:

```
// Check whether carry required.  NOTE: a chunk whose target limbs are
// exhausted has a zero-width left-hand side (SelectExact yields the
// zero register) which cannot absorb any bits.  Such a chunk instead
// asserts its right-hand side is zero, forcing the surplus source limbs
// — and any carry into them — to zero.  Threading a carry out of it
// would push the chunk's entire value into that carry, whose width then
// grows beyond the field's register width; this is the same blowup
// described for borrows in Subtraction.
if lhs > 0 && lhs < rhs && i+1 < len(chunks) {
```

This basically only adds the carry line for a non-zero `lhs` and when we are not on the final chunk.

DavePearce:
A subtle bug existed in the register splitting algorithms for addition and concat which could cause a panic.  The scenariou occurred when there were too few target limbs, and a carry was created that was too large.

This also tweaks the lowering of division as per the TODO item listed against #2117.


pkg/zkc/vm/internal/transform/lower_division.go
```diff
@@ -122,7 +122,7 @@ func expandDivRem[W word.Word[W]](q, r, w, x bytecode.RegisterId, y bytecode.Ope
 		// TODO: rw1 should be of nY width, but because of
```
letypequividelespoubelles: Comment to rm

letypequividelespoubelles (changes requested):

pkg/zkc/vm/internal/transform/lower_division.go
DavePearce: Oops

DavePearce: Fixed now

letypequividelespoubelles (approved):
LGTM

PR #2130, opened 2026-08-11
perf(constraints): dispatch transformation also when all cases sharing same other condition

### LINKED ISSUE #2121: Functiun call after and outside switch
Consider the following:
```
pub input data(address:u16) -> (byte:u8)

fn main() -> (){
    var x:u8 = data[0]
    var v:u8
    var r:u8
    switch x {
        case 0: {v=0}
        case 1: {v=1}
        default: {fail}
    }

    if v == 0 {r = test(x)} 
    else {r = 0}
}

fn test(x:u8) -> (y:u8){
    y = x
}
```

Without this PR, the boolean (lookup selector) is defined as:
```
(vanish main:lookup_sel_10#0 (
   +
   (* (
      +
      (* $ret $b_switch_1_case_1 (- 1 $4))
      (* $ret $b_switch_1_case_2 (- 1 $4))) (- 1 $lookup_sel_10))
   (* (+ $4 (- 1 $ret) (- 1 $b_switch_1_case_1)) (
      +
      $4
      (- 1 $ret)
      (- 1 $b_switch_1_case_2)) $lookup_sel_10)))
```

Image

with this PR:

Image

```
(vanish main:lookup_sel_10#1 (
   *
   (+ $4 $b_switch_1_case_default (- 1 $ret))
   $lookup_sel_10))
```

THe more case you have in the switch, the bigger the effect will be.

IRL, It reduces the max degree of process B_type from 7 to 4.

letypequividelespoubelles:


DavePearce (commented):
At this stage, I cannot actually understand what this algorithm is doing.  I couldn't find a small test case which demonstrated it doing anything new.  I think `rewriteComplementDisjuncts` needs some more documentation about what its doing.  Roughly I understand that we have e.g. `(b0 != 0 && y == 0) || (b1 != 0 && y == 0)` we want to transform into `b_default == 0 && y == 0` (I think).

DavePearce (approved):
LGTM.  In general, it would be good when we rework the branch table optimisation system to provide a general purpose mechanism for describing rewrites like this, such that each rewrite has some structure, etc.

PR #2133, opened 2026-08-11
fix: adjust complexity metric to account for number of columns in constraint

### LINKED ISSUE #2132: fix: adjust complexity metric to account for number of columns in constraint
@ThomasPiellard suggested the following measurement for complexity

```
per column cost for trace commitment with S = size of columns
->  O(S · log S) for the RS-encode each poly (iFFT of size S + coset FFT of size r·S)  part
->  O(S)  for the Merkle tree over r·S / 2 paired leaves, each hashing 2N field elements part
This phase is heavy, the big O notation hides the base cost of a hash, which might be big (e.g. with poseidon2)

per constraint cost of the AIR quotient computation with D = degree of constraint, N = number of distinct columns involved
-> O(D · S · log(D·S)) for the Coset-extend from Lagrange-S to coset-Lagrange
  -> paid N times
-> O(D · S · |DAG|)  (≈ O(N·D²·S) for a generic degree-D relation)) for the 'Evaluate vanishing DAG at D·S points' part
-> O(D · S · log(D·S)) for the pointwise divide by Z_H, coset-Lagrange → canonical iFFT of length D·S
-> O(D · S · log S)  for the per-chunk size-S FFT × D chunks part
There is no hidden cost behind the 'O' notation, the operations that happen at that stage are finite field operations (+,* etc)

for the PCS open part:
-> O((N+D) · S) for the barycentric evaluate N + D polys at zeta·ω^k (few shifts each)
-> O((N+D) · S) for the DEEP quotient RLC over N + D polys on the size-r·S codeword part
-> O(S) for the FRI commit (the low degree test only)
```

Our current complexity metric sums `D²` once per constraint in a given module, which doesn't account for `N`, the number of distinct columns that partake in the constraint. So we measure $\sum d^2$ when we should be measuring  $\sum n_d \cdot d^2$ with $n_d$ the number of distinct columns involved in the constraint (shifts are considered extra columns in this accounting).

**Note.** `S` cannot be known in advance but we may be able at a future point to provide relative weights of the modules

OlivierBBB:
This PR
- changes the complexity metric from `d^2` per constraint of degree d to `n*d^2` when the constraint involves `n` unique cells
- running `zkc compile --stats -v path/to/file.zkc` prints 2 extra matrices that document what `(n,d)` pairs contribute most to overall complexity


letypequividelespoubelles (dismissed):
Can you change the column description in what's printed to add an as well please ?

DavePearce (approved):
LGTM

PR #2135, opened 2026-08-12
perf: fail case in dfa

### LINKED ISSUE #2131: perf: `fail` branch condition improvment
Consider the following:

```
fn test(cond:u1,  x:u8) -> (r:u8){
    if (cond == 1) {fail}
   
    if (x == 0) {r = 0} 
    else {r = 1} 
}
```

it produces the constraint `(vanish test:inst#3 (* $ret (- 1 cond) $3 (- r 1)))`

But it could be simplify to `(vanish test:inst#3 (* $ret $3 (- r 1)))`

Indeed, when we reach the `x==0` case it is not needed to guard by `cond == 0`: a trace with `cond = 0` would be rejected by the previous constraint (from the `fail`.

letypequividelespoubelles:


pkg/zkc/util/dfa/path.go
```diff
@@ -24,39 +24,65 @@ type PathTransferFunction[W util.BigInter, I any] func(offset uint, code I, stat
...
+	return Path[W]{
+		logical.Truth[BranchId, BranchEquality](true),
+		logical.Truth[BranchId, BranchEquality](false),
+	}
 }
 
 // Path adapts a branch condition to be an instance of State.
 type Path[W util.BigInter] struct {
 	condition BranchCondition
+	// dead holds the condition under which this code is reached only through
```
DavePearce, line 36: The comment doesn't really make sense, since you cannot reach a statement through a fail.  Its more like you reach a statement by avoiding a fail ?

DavePearce (dismissed):
Overall, this is fine to merge.  I think the idea that `fail` instructions should not increase the branch condition for downstream instructions is a good one.  However, I'm not a huge fan of this PR as it because it is basically a hack to work around an existing limitation of the branch table optimiser.  To understand this, we consider this function:

```
fn f(x:u8) -> (r:u8) {
  if x == 0 {
    fail
  }
  //
  r = 1
}
```

Translating this with the current change gives this:

```
(vanish f:inst
   (if (!= $ret 0)
      (∧
         (if (== x 0)
            (== $2 0))
         (if (!= x 0)
            (== $2 1)) 
         (!= $2 0) 
         (== r 1))))
```

Now, translating with the current changes gives the following:

```
(vanish f:inst
   (if (!= $ret 0)
      (∧
         (if (== x 0)
            (== $2 0))
         (if (!= x 0)
            (== $2 1)) 
         (!= $2 0)
         (if (!= $2 0)
            (== r 1)))))
```
What we can see is that, in fact, the problem is that the simplifier does not automatically reduce `(if (!= $2 0) (== r 1))` to just `(== r 1)` even though it would be valid for it to do this.  **Hence, the problem is really in the simplifier itself and not anywhere else**.  Since this PR does not address that, it is not actually attacking the right problem.  Rather, it is hiding the problem with yet more code.

Anyway, I'm happy to merge since the branch condition optimisation process will be reworked soon enough, and when that happens this will be addressed properly.

DavePearce (approved):
Ok, after some surprisingly extensive analysis ... LGTM.

PR #2137, opened 2026-08-12
feat: remove HIR representation (second attempt)

### LINKED ISSUE #2124: feat: remove HIR representation (Part 2)
# Summary

The goal of this PR is to continue removing unnecessary syntax from the Corset language, and ideally remove HIR altogether.

DavePearce:
Remove the HIR layer entirely: the Corset compiler now translates
straight from the AST into MIR. Since MIR arithmetic terms cannot hold
conditionals, the translator lifts `(if c a b)` out of arithmetic
position by case splitting (see pkg/corset/compiler/conditional.go),
fusing what were previously two passes into one — the remainder of the
old lowering was structural copying and simply disappears. Generated MIR
and AIR are unchanged for every valid and agnostic fixture. This also
removes the binary file format (pkg/binfile) together with the `compile`
command and `.bin` inputs, so constraints are always compiled from
source, and with it the gob encoders that existed to serve it (HIR,
schema.Table, UniformSchema, register.Refs); the `generate` command
emitting Java trace classes goes too. Finally, a number of Corset source
constructs are retired: casts and maskwidth, `^` exponents, `::`
concatenation, explicit normalisation, and labelled/externalised
constants, with definrange narrowed from an arbitrary expression to a
column.


letypequividelespoubelles (approved):
blind approval

PR #2138, opened 2026-08-13
feat: variable timestamp width for read-write memory

### LINKED ISSUE #2069: feat: variable timestamp width for read-write-memory

OlivierBBB:
## Summary

Implements the first half of #2069: the timestamp of a read-write memory is now a **mandatory, per-memory declaration** instead of a hardcoded 32 bits:

```zkc
memory data[u16](addr:u16) -> (byte:u8)

type WideTimestamps = u48
memory ram[WideTimestamps](addr:u32) -> (word:u64)   // aliases resolve
```

The bracketed type must resolve to a fixed-width `uN` with 1 ≤ N ≤ 64 (the runtime clock path is uint64). The width is a **capacity cap** — it bounds the number of accesses per execution (overflow fails at trace-generation time via the check-cast on `stamp+1`) — not a per-row cost knob. `input`/`output`/`static` memories take no timestamp type.

The second half of #2069 (replacing the carry-based monotonicity gadget with `@k` prefix-equality bits + one `TIMESTAMP_DELTA` column) is deliberately left for a follow-up PR.

## How it flows

1. parser → AST (`decl.Memory.TimestampType`, a `data.Type[S]`) → linker (`linkType`) → validator (`typeMemory`: uN, closed, ≤64, RAM-only);
2. codegen extracts the width into the VM descriptor (`descriptor.Memory.TimestampWidth()`), which the mid-pipeline rebuild sites (register splitting, word conversion) forward and the binfile gob carries;
3. the three previously hardcoded `stampWidth = 32` constants (timestamp threader, RAM constraint generator, trace observer) now read the descriptor — per effect in the threader, so two RAMs of different widths can share a row.

## Notes for review

- **Binfile**: minor version bumped to 1. Decode is tolerant: pre-#2069 files hit `io.EOF` on the width field and default read-write memories to the then-implicit 32, preserving the "lower minor stays readable" contract.
- **Breaking (source level)**: every `memory` declaration needs `[uN]`. All in-tree fixtures migrated with `[u32]`; the linea-monorepo riscv corpus needs the same one-line-per-memory change before bumping its zkc version.
- Verified: `[u32]` vs `[u48]` changes a RAM module from 15 to 19 columns on `KOALABEAR_16` (one extra limb per timestamp family + one carry); widths survive a `.bin` round-trip; ram_13 exercises two different widths on one trace row.
- New negative fixtures memory_17..21 pin the diagnostics (missing width, `u65`, `𝔽`, array type, width on `pub input`).

## Testing

- `make lint` / gofmt / `make zkc-lint`: clean.
- Full `Test_ZkcUnit|Test_ZkcMixed|Test_ZkcInvalid` suite: green (incl. new ram_12/ram_13, memory_17..21).
- `make unit-test` and the gogen suite: green.

Part of #2069.


pkg/zkc/compiler/validate/typing.go
```diff
@@ -218,7 +218,7 @@ func (p *TypeChecker) typeFunction(fn decl.ResolvedFunction) []source.SyntaxErro
 	return errors
 }
 
-func (p *TypeChecker) typeMemory(c decl.ResolvedMemory) []source.SyntaxError {
+func (p *TypeChecker) typeMemory(c *decl.ResolvedMemory) []source.SyntaxError {
```
OlivierBBB, line 223: Due to, I guess, the inclusion of `c` in a syntax error.

pkg/zkc/compiler/parser/parser.go
```diff
@@ -502,6 +502,11 @@ func (p *Parser) parseInputOutputMemory() (decl.Unresolved, []source.SyntaxError
 	if name, errs = p.parseIdentifier(); len(errs) > 0 {
 		return nil, errs
 	}
+	// A timestamp type is only meaningful on read-write memory, which has a
+	// timestamp to size; reject it here with a purposeful message.
+	if p.lookahead().Kind == LSQUARE {
+		return nil, p.syntaxErrors(p.lookahead(), "only read-write memory takes a timestamp type")
+	}
```
OlivierBBB, line 509: Strictly speaking unnecessary.

pkg/zkc/compiler/ast/decl/memory.go
```diff
@@ -44,6 +46,27 @@ const (
...
+	case PRIVATE_STATIC_MEMORY, PUBLIC_STATIC_MEMORY:
+		return true
+	default:
+		return false
+	}
+}
+
+// RequiresTimestampType reports whether this memory kind carries a timestamp
+// type (true only for read-write memory).
+func (kind MemoryKind) RequiresTimestampType() bool {
```
DavePearce: Not sure this is the best name, since "timestamp type" is not something which really makes sense here.  Perhaps "timestamp width" but for me `HasTimestamp()` makes the most sense to me (i.e. because having a timestamp always implies it must have a width).

pkg/zkc/vm/internal/descriptor/memory.go
```diff
@@ -25,23 +26,25 @@ import (
...
-// only).  NOTE: this will panic is a non-static memory is created with some
-// contents.
-func NewMemory[W word.Word[W]](name string, registers []Register[W], kind MemoryKind,
-	contents []W) *Memory[W] {
+// initializes a memory region with a given name, kind (ROM, RAM, WOM),
+// timestamp width (for read-write memories only, zero otherwise), registers
+// (whose geometry gives size and layout) and initial contents (for static
+// memories only).
+// NOTE: this will panic is a non-static memory is created with some contents.
+func NewMemory[W word.Word[W]](name string, kind MemoryKind, timestampWidth uint,
```
DavePearce: To be honest, I would change this from `timestampWidth uint` to `timestampWidth util.Option[uint]` and then add a sanity check below that `timestampWidth.HasValue() <==> kind.RequiresTimestampType()`.

pkg/zkc/vm/internal/descriptor/memory.go
```diff
@@ -195,6 +208,16 @@ func (p *Memory[W]) GobDecode(data []byte) error {
 	if err := gobDecoder.Decode(&p.contents); err != nil {
 		return err
 	}
+	// The timestamp width was added in binfile v0.1 (issue #2069); files
+	// predating it simply end here, in which case read-write memories get the
+	// then-implicit width of 32.
+	if err := gobDecoder.Decode(&p.timestampWidth); err == io.EOF {
```
DavePearce: No, don't support legacy binfile format --- this is unnecessary, and potentially could hide issues.  We can just assume all binary files are compiled to the latest format.

pkg/zkc/vm/internal/interpreter/memory_log.go
```diff
@@ -98,12 +98,12 @@ type TraceableMemoryLog[W word.Word[W]] struct {
 
 // Read implementation for Log.
 // valueWritten will coincide with the read value.
-func (l *TraceableMemoryLog[W]) Read(address uint64, timestampWritten uint64, valueWritten W) {
+func (l *TraceableMemoryLog[W]) Read(address uint64, valueWritten W, timestampWritten uint64) {
```
DavePearce: Not sure why the ordering changed here, since this is just going to cause conflicts with other PRs.  Eitherway, its fine if that is preferred.

DavePearce (dismissed):
Basically, this looks good.  I left a few minor comments only.

pkg/zkc/vm/internal/interpreter/memory_log.go
OlivierBBB: I asked Claude to ensure that arguments were always in the same order: address, value, timestamp.

DavePearce (dismissed):
LGTM

DavePearce (approved):
Second approval.

PR #2141, opened 2026-08-13
2119 rewriting a registerszeroregister shouldnt trigger a new vector during vectorization

### LINKED ISSUE #2119: Rewriting a `registers.ZeroRegister()` shouldn't trigger a new vector during vectorization
Allocating a ZeroRegisters, and then having an assignement on it is considered as a second write for the vectorization, thus making a MLI fct:

For exemple, try this is lowerDivision, and any /, % or /% operation will turn a MLI.

```
z   = registers.ZeroRegister()
subZ1 = bytecode.SubConst(z, []bytecode.RegisterId{rw1}, y.AsConstant())
```

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/split/allocator.go
```diff
@@ -90,6 +90,10 @@ func (p *registerAllocator[W]) Allocate(prefix string, width util.Option[uint])
 		// Default padding (for now)
 		zero W
 	)
+	// Reuse any existing zero register.
+	if width.HasValue() && width.Unwrap() == 0 {
```
DavePearce, line 94: haha, sneaky

pkg/zkc/vm/internal/transform/split/allocator.go
```diff
@@ -141,10 +145,10 @@ func (p *registerAllocator[W]) Width() uint {
...
 	// Check for any existing zero registers;
 	for i, r := range p.registers {
-		if r.Bitwidth().UnwrapOr(math.MaxUint) == 0 {
+		if bytecode.IsZeroWidth(r) {
 			return util.Cast[RegisterId](uint(i))
 		}
 	}
 	// Allocate a new register.
-	return p.Allocate("z", util.Some[uint](0))
+	return p.AllocateNamed("cst_0", util.Some[uint](0))
```
DavePearce, line 153: What does `cst` stand for???  I guess constant?  Maybe just `const_0` ?  Anyone doesn't matter.

DavePearce (approved):
LGTM.  I think `cst_0` is not the best name of all time, but its fine.  I probably would have used `const_0`.  Don't change it though.

pkg/zkc/vm/internal/transform/split/allocator.go
letypequividelespoubelles: yeah, just to make it shorter; I ddidn't want 0, as for other constant, I want to have like `$cst_XX` name; But can change for sct / const / I really don't care

PR #2142, opened 2026-08-13
feat: optimise lookup constraint checking

### LINKED ISSUE #2085: feat: optimise constraint checking for lookups
# Summary

Currently, lookups are checked one-by-one with no shared data between them.  This is actually extremely inefficient, especially when we have a large number of lookups into the same module.  The proposed approach is to cache the `hash.Set` used to store the data for a module and then reuse this across all lookups into that module.  This requires a certain amount of coordination between checking individual constraints --- which is not something that previously has been done.

DavePearce:
This adds a new centralised process called "context construction" which pre-generates the source / target sets required for checking a lookup constraint.  The primary purpose of doing this is to avoid reconstructing the same sets over and over when, for example, we have a large number of lookups into the same module.  The context construct process supports both parallel and sequential construction, but does not attempt to subdivide the construction of a single set (which, in theory, it could).

### DIFF
--- a/pkg/cmd/corset/check.go
+++ b/pkg/cmd/corset/check.go
@@ -248,7 +248,7 @@ func CheckTrace[F field.Element[F]](ir string, schema sc.AnySchema[F], tf lt.Tra
 	//
 	stats = util.NewPerfStats()
 	// Check constraints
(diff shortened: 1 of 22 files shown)


letypequividelespoubelles (approved):
half blind approval

PR #2145, opened 2026-08-13
feat: caller lookups for read-write memory accesses (3/3)

OlivierBBB:
Re-lands #2081, which was accidentally merged into `1807-ram-constraints` instead of `main`. That branch was subsequently rewritten from the ground up and merged separately, so the lookup work never reached `main`. This is a port onto current `main`, not a cherry-pick — the surrounding code moved (see below).

## What this does

Every read-write memory access now emits a lookup tying it to a row of that memory's table, through the same per-site machinery already used for calls and access-once memories. The source tuple carries the accessor's address, data and threaded timestamp registers (gated by the shared line/path selector) and maps onto the table's `ADDRESS`, `VALUE_WRITTEN` and `TIMESTAMP_WRITTEN` columns. A read's data registers are its outputs — the value read back — which the table exposes in `VALUE_WRITTEN` too, since a read row writes back what it found.

The access's read/write kind is pinned by the **target-side filter** rather than a tuple element, because MIR lookup tuples carry register accesses only. The table gains two fully-determined selector columns, `EXEC_WRITE = EXEC * IS_WRITE` and `EXEC_READ` (constrained subtraction-free as `EXEC_WRITE + EXEC_READ == EXEC`), filled by the trace observer. A write site's lookup targets the table filtered by `EXEC_WRITE`, a read site's by `EXEC_READ`.

With this, the guest-facing columns of every access row are pinned by the caller and the row-local constraints hold end-to-end. The remaining gap — that `VALUE_READ` genuinely returns the last value written to the address — is the offline memory-checking bus (#2068).

Part 3 of 3 of the RAM timestamps feature (#1807).

## Differences from #2081

The port is smaller than the original (112 vs 130 insertions) because `main` has since changed underneath it:

- `lookup.FilteredVector` / `NewLookupConstraint` now take `register.Id`s directly instead of term-level register accesses, and `mir.LookupVector` is no longer generic. `emitRamLookup` therefore drops the width plumbing and the `accessorRegs` parameter that the original needed.
- The RAM column-name constants moved from `pkg/asm/io/function.go` (that package is gone) into `pkg/zkc/constraints/trace/builder.go`.
- `pkg/zkc/constraints/post/` → `pkg/zkc/constraints/trace/`, and `determineRamRegisters` → `initReadWriteMemory` using `rtrace.NewColumnDescriptor`. The observer sets the new selectors via the existing `field.Uint1[F]` helper rather than the original's `if`/`else`.

The fixup commit that rode along in #2081 (`fix: restore dispatch-case retargeting dropped by the ours-strategy merge`) is **not** included: its content — the retargeting logic, the `ram_11` fixture and its test — is already on `main` via the rewritten constraints branch.

## Review guide

- `constraints/ram.go` — the two selector columns, their layout slots, and the `exec_write_def` / `exec_read_def` constraints that fully determine them.
- `constraints/call_and_memory_lookup.go` — `emitRamLookup`, plus removal of the `groupLookupsByCondition` skip that used to exclude RAM (the TODO pointing at #1807 is now resolved).
- `constraints/trace/process_ram.go` — the observer filling `EXEC_WRITE` / `EXEC_READ`, and the widened row layout.
- Worth checking: the column order in `computeRamLayout`, `newRamTraceLayout` and `initReadWriteMemory` must stay in lock-step; the new columns are appended last in all three.

## Testing

- `make lint` — 0 issues.
- `make zkc-unit-test` — full suite green (~500s), covering the 11 `ram_*` fixtures across fields, heights and padding strategies.
- `make unit-test` — green.
- Verified the constraints are actually emitted (not a silent no-op) via `zkc compile --mir testdata/zkc/unit/ram_02.zkc`: read sites filter on `ram:$exec_read`, write sites on `ram:$exec_write`, both targeting `(address, byte, $ts_written_*)`.

### DIFF
--- a/pkg/zkc/constraints/call_and_memory_lookup.go
+++ b/pkg/zkc/constraints/call_and_memory_lookup.go
@@ -85,8 +85,12 @@ func addLookups[W vm.Word[W], F field.Element[F]](mod *schema.Table[F, mir.Const
 					emitCallLookup(mod, ctx, callerRegs, uint(pc), uint(c.Target),
 						toRegisterIds(c.Arguments), toRegisterIds(c.Returns), srcSelector, infos)
 				case *vm.BytecodeReadWrite[W]:
-					emitMemoryLookup(mod, ctx, callerRegs, uint(pc), entry.cc, uint(c.Id),
-						toRegisterIds(c.Address), toRegisterIds(c.Data), srcSelector, infos)
+					if infos[c.Id].(*vm.Memory[W]).IsReadWrite() {
+						emitRamLookup(mod, ctx, uint(pc), entry.cc, c, srcSelector, infos, field)
+					} else {
+						emitMemoryLookup(mod, ctx, callerRegs, uint(pc), entry.cc, uint(c.Id),
+							toRegisterIds(c.Address), toRegisterIds(c.Data), srcSelector, infos)
+					}
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


DavePearce (approved):
Yeah, LGTM

PR #2149, opened 2026-08-14
chore: fix check on max static height

### LINKED ISSUE #2086: bug in maxStaticHeigth check
Actually, there is a check that `maxStaticHeigth>=2` (reminissence of first range lookup check, now useless). 

The real compile check should be `maxStaticHeigth>= max heighth of declared static module` 

_Originally posted by @letypequividelespoubelles in https://github.com/LFDT-Lineth/zkc/pull/2061#discussion_r3687915745_

letypequividelespoubelles:


pkg/test/util/util.go
```diff
@@ -48,17 +48,24 @@ type TestCase struct {
 // CompileMachine compiles one or more zkc source files into a base machine for
 // executing tests with.
 func CompileMachine(field field.Config, srcfiles ...source.File) []source.SyntaxError {
```
DavePearce: I think `CompileMachine` can be removed in fact ... its not used.

pkg/test/util/util.go
```diff
@@ -48,17 +48,24 @@ type TestCase struct {
...
 // This includes both the validation phase and the code generation phase.
 func CompileZkc(field field.Config, srcfile source.File) []source.SyntaxError {
-	program, _, errors := compiler.Compile(field, srcfile)
+	return CompileZkcWith(codegen.DEFAULT_CONFIG.Field(field), srcfile)
+}
+
+// CompileZkcWith compiles a single zkc source file under a given codegen
+// configuration, potentially producing errors.  This includes both the
+// validation phase and the code generation phase.
+func CompileZkcWith(config codegen.Config, srcfile source.File) []source.SyntaxError {
```
DavePearce, line 57: <strike>No, this doesn't make sense to me.  Just remove the old `CompileZkc()` above and then rename this one to `CompileZkc()`.</strike>

DavePearce: Ok, no worries --- I see what you are doing.

pkg/zkc/vm/internal/transform/lower_or_xor_and.go
```diff
@@ -42,10 +42,15 @@ func LowerOrXorAnd[W word.Word[W]](program descriptor.Program[W]) descriptor.Pro
 		out = slices.Clone(program.Modules())
 		// maxStaticWidth is floor(log2(maxStaticHeight)); a bitwise table indexes
 		// two w-bit operands, so it fits only when 2w <= maxStaticWidth.
-		bitwiseStaticWidth = uint(bits.Len(program.MaxStaticHeight())-1) / 2
-		helpers            = newBitwiseHelpers[W](uint(len(out)), bitwiseStaticWidth)
+		bitwiseStaticWidth = func() uint {
```
DavePearce: Can I suggest we have a function `Program.MaxStaticWidth()` maybe?

pkg/zkc/vm/bytecode.go
```diff
@@ -170,17 +170,7 @@ func ValidateProgram[W word.Word[W]](p Program[W]) error {
 
 // NewBytecodeProgram assembles a bytecode program directly from pre-lowered
 // descriptor modules, bypassing the word-machine round trip.
-//
-// NOTE: maxStaticHeight must be non-zero.  Downstream passes derive the maximum
-// static table width from it as floor(log2(maxStaticHeight)), computed as
-// bits.Len(maxStaticHeight)-1.  A zero height silently underflows that
-// expression to the maximum uint, whereupon every register width is treated as
-// statically enumerable and table generation exhausts memory.
 func NewBytecodeProgram[W word.Word[W]](field field.Config, maxStaticHeight uint, modules ...Module[W]) Program[W] {
```
DavePearce, line 182: Why remove the check that `maxStaticHeight` is non-zero ?

DavePearce (dismissed):
Basically, looks good to me.  I didn't understand why you removed the check that `maxStaticHeight != 0`, as this then makes some other functions a bit messier (i.e. because they handle `maxStaticHeight == 0` explicitly).

pkg/zkc/vm/internal/transform/lower_or_xor_and.go
DavePearce: I'm not actually sure that is a good idea.  I think it would just be better to be able to assume that `maxStaticHeight != 0` ?

pkg/test/util/util.go
letypequividelespoubelles: deleted

DavePearce (dismissed):
LGTM

DavePearce (approved):

PR #2152, opened 2026-08-14
chore: add check for unreachable module

### LINKED ISSUE #2016: prune (or warn) for unreachable fct at compile time.
Specifically, the function `Prune()` in `descriptor.Program` is intended (once implemented) to remove all functions which cannot be reached from the `main` entry point.  Thus, this would also include `#[debug]` functions when `quiet` was enabled as they would not be reachable.

@letypequividelespoubelles iho, this shouldn't be a silent drop, but a compile warn that fct are unreachable.

_Originally posted by @DavePearce in https://github.com/LFDT-Lineth/zkc/pull/2012#pullrequestreview-4739097269_

letypequividelespoubelles:


pkg/zkc/vm/validation.go
```diff
@@ -64,11 +64,6 @@ func validateBytecodeProgram[W word.Word[W]](program Program[W]) error {
 				mid, module.Name(), zeros))
 		}
 	}
-	// An environment stores its enclosing module as a uint16.  If the module
```
letypequividelespoubelles, line 67: this was a duplicate, already existing line 31-34

DavePearce (approved):
Yeah, looks good.  I think we would want a slightly more generic way of construct the graph from the constraints, especially once we add the bus primitives.  But, for now, seems good.

PR #2166, opened 2026-08-19
perf: relax multiplication splitting granularity

gbotrel:
Two changes to the schoolbook multiplication splitter: a **larger sub-limb granularity** and a **fusion of the final accumulation with the target assignment**. Together they roughly halve the post-split register count of multiplication-heavy code.

## 1. Relax the granularity rule

The splitter picked its sub-limb width `g` as the largest **divisor** of `RegisterWidth` with `2g <= BandWidth`, forcing g=8 on `KOALABEAR_16` — 64 partial products for a u64 multiply.

This relaxes the divisibility rule: sub-limbs are cut on a grid independent of the limb layout (a boundary crossing costs one destructure + one concat), so `g` reaches `min(RegisterWidth, BandWidth/2)` — 15 on KoalaBear, 25 partial products. Granularity is additionally bounded so column accumulation provably cannot wrap the field modulus (`g + bitlen(T+1) <= BandWidth`, with `T` the worst-case column fan-in), falling back to the legacy divisor rule where unattainable (tiny test fields) — no existing program is rejected. Aligned grids (e.g. the fast-mode word configs) emit exactly the same bytecode as before.

The core is plain schoolbook long multiplication over g-bit sub-limbs — each partial product is at most 2g bits, landing in two adjacent weight columns, and each column is one add whose carry feeds the column above:

```
                        a₁     a₀
             ×          b₁     b₀
             ─────────────────────
                     a₀b₀ʰ  a₀b₀ˡ      one narrow mul per pair;
              a₁b₀ʰ  a₁b₀ˡ             ˡ/ʰ = low/high g bits, at
              a₀b₁ʰ  a₀b₁ˡ             column weight i+j / i+j+1
       a₁b₁ʰ  a₁b₁ˡ
             ─────────────────────
out₃ ⟵c₂⟵ out₂ ⟵c₁⟵ out₁    out₀       one add per column, carries
                                       threaded into the next column
```

## 2. Fuse the final fold with the target assignment

The first version of this PR produced the product on its own g-grid and then reshaped it into the target:

```
naive:  accumulate columns → clamp each column to the target width
        → re-cut at the target-limb boundaries → concatenate
        (a mov chain per column)
```

Now the final fold's columns are laid **straight onto the target grid**: each column add writes output pieces already clamped to the target width and already cut at the target-limb boundaries, and a column that is already identical to its output (a single partial of exactly the column width) is passed through with no instruction at all. For `u16 × u16 → u32` with g = 15:

```
  col 2       col 1            col 0
|── 2 ──|─────── 15 ──────|─────── 15 ───────|  product columns
   m₂       p₁       p₀      (pass-through)     add c::p₁::p₀ = …
|── 2 ──|── 14 ───|─ 1 ──|─────── 15 ───────|   add m₂ = …
└────── hi ───────┘└────────── lo ──────────┘   one cat per limb
```

Concretely, `hi::lo = x as u32 * y as u32` on KoalaBear went from 20 registers / 15 instructions to 15 / 10:

```
mov s$5::s$4 = x                          add c$17::m$16::m$15 = pp$9+pp$10+pp$12
mov s$7::s$6 = y                          add m$18 = pp$11+pp$13+pp$14+c$17
mul pp$9::pp$8   = s$4*s$6                cat lo = m$15::pp$8
mul pp$11::pp$10 = s$4*s$7                cat hi = m$18::m$16
mul pp$13::pp$12 = s$5*s$6                ret
mul pp$14        = s$5*s$7
```

Product bits at or beyond the target width fall into zero-width columns whose adds are satisfiable only by zero — exactly the overflow check the unsplit multiply performs. A clamped column (width < g) deliberately gets **no carry**: for any in-range product its sum provably fits the clamped width, so letting its add reject overflow directly *is* the overflow check, and it keeps every carry at its small full-granularity size.

## Results

Post-split registers / AIR complexity on `KOALABEAR_16`:

| program | main | +granularity | +fusion |
|---|---|---|---|
| u64 × u64 → u128 | 207 | — | 110 (−47%) |
| u64 divmod | 434 | — | 232 (−47%) |
| bench: modexp32 | 475 / 22021 | 378 / 18087 | 325 / 15596 |
| bench: fast_pow | 150 / 6992 | 116 / 5149 | 96 / 4375 |
| bench: gcd | 101 / 3756 | 84 / 2909 | 74 / 2603 |
| bench: div_rem | 338 / 15377 | 304 / 13683 | 284 / 13071 |
| bench: fnv1a_hash | 258 / 6063 | 224 / 5203 | 210 / 4871 |
| bench: keccak | 3712 / 203132 | 3713 / 203608 | 3659 / 195360 |

`GF_8209` shows the same trend (modexp32 1257 → 834, fast_pow 419 → 259); `BLS12_377` is unchanged (no splitting). The granularity change alone saved **618 columns in the R5 interpreter in linea-monorepo**; division benefits for free since `LowerDivisions` lowers to a multiply.

## Validation

Beyond the test suite, the transform was ported function-by-function to a Python model (verified to reproduce the compiler's output instruction-for-instruction) and checked over ~11k configurations / 852k executions across register-width/bandwidth combinations, operand counts, constants and target widths:

- every in-range product executes to the exact value, and every out-of-range product is rejected;
- for every emitted instruction, the statically-maximal RHS fits the field bandwidth whenever the new bound (rather than the legacy fallback) chose `g` — so no constraint can wrap the modulus, and since each multi-target add is a range-checked mixed-radix decomposition, the witness is uniquely determined;
- over-wide register allocations occur only on legacy-fallback tiny-field configs, where `main` panics identically today (the fusion actually reduces their incidence).

All ZkC unit/mixed/invalid, util and bench tests pass on all fields. The final commit is a pure readability refactor, verified byte-identical over all 1424 fixture compilations (every zkc fixture × four fields).


DavePearce (approved):
LGTM

PR #2174, opened 2026-08-24
feat: support no return functions

### LINKED ISSUE #2173: feat: support "no return" functions
_(see also #1811)_

# Summary

A no return function is one which is guaranteed never to return (e.g. because it always hits a `fail` statement).  Supporting "no return" functions may seem a little odd, but they essentially provide the key mechanism needed for splitting execution across shards.  For efficient implementation, it seems likely we additionally would require #1811.

## Syntax

The syntax for a "no return" function follows Rust in using the so-called ["never" type](https://doc.rust-lang.org/std/primitive.never.html).  For example, it would look like this:

```
fn f(x:uint32) -> ! {
  fail
}
```

This is a minimal example of a function which never returns.  A key challenge with "no return" functions is that we additionally require some way to indicate successful termination.  Therefore, we introduce syntax specifically for this, like so:

```
fn f(x:uint32) -> ! {
  done
}
```

DavePearce:
This adds syntax for so-called "no return" functions, along with the corresponding `done` statement.


letypequividelespoubelles (approved):
blind approval

PR #2177, opened 2026-08-24
fix: support wide concatenation sources in gogen

### LINKED ISSUE #2167: fix: `gogen` fails for wide concatenation source
# Summary

On some programs, `gogen` currently fails with an error `ERRO[0000] gogen: concatenation source wider than 64 bits unsupported`

## Example

A minimal example is:

```
fn main() {
  mul64_ss(1,1)
}

fn mul64_ss(lhs:u64, rhs:u64) -> (hi:u64, lo:u64) {
    var val:u128 = (lhs as u128) * (rhs as u128)
    hi::lo = val
}
```

Then, the following command:

```
> zkc exec -f --gogen test.json test.zkc
```

fails with the following error:

```
ERRO[0000] gogen: concatenation source wider than 64 bits unsupported
exit status 4
```

## Notes

The interesting thing is that this variation is fine:

```
fn mul64_ss(lhs:u64, rhs:u64) -> (hi:u64, lo:u64) {
    hi::lo = (lhs as u128) * (rhs as u128)
}
```

gbotrel:
Fixes #2167.

`emitConcat` rejected any CAT source represented as a lo/hi pair, so `hi::lo = val` with `val:u128` failed under `--gogen`. The interpreter has no such restriction.

- Single-source CAT is now a plain store via `storeValue` (also uses the tighter interval bound).
- The multi-source packing loop handles wide sources at any position.
- Regression test: `basic_88` (the issue's example, self-checking against the interpreter via the gogen cross-check).


DavePearce (approved):
LGTM

PR #2179, opened 2026-08-24
feat: support CLI options for compile command

### LINKED ISSUE #2178: feat: cli options for `compile` command
# Summary

This adds some command-line options for the `compile` command to help debugging its output. For example, by allowing ANSI escapes to be disabled, etc.

DavePearce:
This puts through two specific CLI options: (1) to disable ANSI escapes upon request; (2) to prevent "skip arrows" from being shown upon request.  These are really just to assist with controlling the output for debugging / analysis reasons.

More details (from Claude): Small quality-of-life PR: two new flags to tame the
  tool's output for debugging and analysis (e.g. piping to a file or diffing).

  1. --ansi-escapes (global, default true). Set it to false to strip
     colour/escape codes¹ from all output.
     - 1.a A package-level AnsiEscapes variable in root.go, filled in
       once in a PersistentPreRunE hook so every subcommand sees it.
     - 1.b All the hard-coded tbl.Print(true) call sites (compile, stats,
       trace) and the constraint-check report in execute.go now pass that
       variable instead of true.
  2. --show-skips (compile command, default true). Set to false to drop the
     "skip arrows" — the drawn flow-graph column showing where conditional
     skips jump to — from the bytecode listing. When off, each row just gets
     an empty string where the arrows would be.

  Both default to on, so plain invocations behave exactly as before; you
  opt out with --ansi-escapes=false / --show-skips=false.

### DIFF
--- a/pkg/cmd/zkc/compile.go
+++ b/pkg/cmd/zkc/compile.go
@@ -80,6 +80,8 @@ type CompileConfig struct {
 	// order determines how modules are ordered in the --stats output
 	// (name|total|complexity|lookups).
 	order string
+	// indicates whether or not to print skip arrows
+	skipArrows bool
 	// indicates whether or not to print everything in full (e.g. including
 	// static reference tables).
 	verbose bool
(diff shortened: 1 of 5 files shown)


OlivierBBB (approved):
LGTM

PR #2181, opened 2026-08-25
feat: lisp syntax and sharded trace checking for send and receive bus primitives

### LINKED ISSUE #1722: feat: support bus `send` / `receive` primitives
## Summary

A desired feature for `go-corset` is to support a notion of a "logup" bus or similar.   This would operate using `send` and `receive` primitives which operate over columns (much as lookups already do).

## Example Syntax

Here is some imaginary syntax expressed in the old `corset` style.  Whilst we will not have syntax like this, it can be helpful to view it in this way.

```
(defconst MSG_TYPE 123)
(defcolumns (A :i32) (B:i64) (C:i64))
(defsend s1 MSG_TYPE (A B C ...))
```

```
(defcolumns (X:i32) (Y:i64) (C:i64))
(defreceive r1 MSG_TYPE (X Y Z))
```

Some additional features we would desire include:

* (**Selector Column**) This column acts like the selector in a conditional lookup.  Specifically, it is a binary activation column which actives the send/receive when it holds `1` and deactivates it otherwise.  This means, for example, that the selector column can be `0` in the padding region.  We might designate the selector column like this (where `X` is the binary selector column):
```
(defsend s1 MSG_TYPE X (A B C ...))
```

* (**Multiplicity Column**) This is an (optional) column which can hold a "multiplicity" value.  When the column is not given, the multiplicity defaults to `1`.  Otherwise, the value of the multiplicity column indicates how many messages are sent / received.  For example, a multiplicity of `2` would indicate an identical message sent/received twice.

## Notes

* (**Domain**) An interesting question is whether or not a send or recieve can be disabled on the first (or last) row of the entire trace.  That is, when the trace is divided across multiple machines, it remains the case that there is never more than one first and one last row (though they could be the same row in principle, or not even exist).  This analguous to the idea of "domain" of vanishing constraints, which can be enabled only for the first or last row of a trace (a.k.a local constraints).

* (**Multiplicity Overflow**) Theoretically speaking, it may be possible for the multiplicity column to overflow.  In practice, this is probably unlikely --- or even impossible --- to happen.  Nevertheless, should this happen, then tracing should just fail with an error for now.

OlivierBBB:


pkg/schema/constraint/bus/constraint.go
```diff
@@ -0,0 +1,267 @@
...
+func (p Constraint[F]) accumulate(tr trace.Trace[F], ports []Port, tally *Tally[F], sign int) {
+	for _, port := range ports {
+		var trModule = tr.Module(port.Module)
+		//
+		for row := range trModule.Height() {
+			if trModule.Column(port.Selector.Unwrap()).Get(row).IsZero() {
+				continue
+			}
+			//
+			var message = make([]F, port.Len())
```
DavePearce, line 194: FYI --- there is some inefficiency here in terms of memory allocation.

DavePearce (approved):
Overall, a great job.

pkg/schema/constraint/bus/constraint.go
OlivierBBB: Thanks, I'll have Claude take a look. There are other places that are inefficient (e.g. verification of the bus constraints are attempted too often and in cases where they are doomed to fail, e.g. on individual shards).

OlivierBBB: I'll put them in a follow-up PR

PR #2185, opened 2026-08-25
perf: skip call to AND/OR/XOR for specific constant right operand

### LINKED ISSUE #2009: perf: AND / OR / XOR with absorbing / null constant shouldn't trigger and / or / xor call
Let's have the code 

```
func (x:u32) -> (y:u16) {
y = (x & 0xFFFF) as u16
}
```

Currently, during codegen, OxFFFF will be assigned to a register `$1`, and then (assuming koalabear_16) we'll have register splitting:
`x = x1::x2`

and then trigger the call to `$bit_and_u16(x2, $1)`.

But we should keep the information of constant till after register splitting. So that we know we'll have `x2 && 0xFFFF` which means we could directly assign `y==x2`. No call to `$bit_and_u16` required for this functiun. 

Same could be done for and, or , xor, skipping calls when dealing with unity / zeroing constant AFTER register splitting.

letypequividelespoubelles:


DavePearce (approved):
Yup, LGTM.  I note that it does seem increase overall complexity, and increases the number of d3 constraints.  But, likewise, decreases the number of registers.

PR #2191, opened 2026-08-26
perf: enable partial function call and static memory read

### LINKED ISSUE #2010: enable partial function call
_(see also #2188)_

Currently we have 3 static tables for AND, OR and XOR. So 3 * (2 + 1) = 9 columns: 3 tables having arg1, arg2, and result. 
We should have only 1 table (arg1, arg2, res_and, res_or, res_xor), saving us 4 columns = 2^18 cells (assuming we have default static tables of depth 2^16).

This requires a "partial" read where instead of reading the triplet we can say which column(s) we want to read from the static tables.

letypequividelespoubelles:


pkg/zkc/compiler/parser/parser.go
```diff
@@ -1975,12 +1975,18 @@ func (p *Parser) parseTupleExpr(env Environment) (Expr, []source.SyntaxError) {
...
-		nexpr Expr
-		errs  []source.SyntaxError
-		name  string
+		nexpr     Expr
+		errs      []source.SyntaxError
+		name      string
+		lookahead = p.lookahead()
 	)
 	//
 	name, errs = p.parseIdentifier()
```
DavePearce: Should this not be just `name, errs = p.parseDeclaredIdentifier()`

pkg/zkc/compiler/parser/parser.go
```diff
@@ -2093,6 +2099,15 @@ func (p *Parser) parseLVal(env Environment) (LVal, []source.SyntaxError) {
 	if len(errs) > 0 {
 		return lv, errs
 	}
+	// "_" as an assignment target discards the corresponding return value of a
+	// function call (or static memory read).
+	if reg == "_" {
```
DavePearce: This should be in the switch, not stuck out on its own.

pkg/zkc/compiler/validate/typing.go
```diff
@@ -300,6 +305,91 @@ func (p *TypeChecker) typeAssignment(s *stmt.Assign[symbol.Resolved], env Variab
...
+// against the source's return types.
+func (p *TypeChecker) typePartialAssignment(s *stmt.Assign[symbol.Resolved], env VariableMap, effects bit.Set,
+) []source.SyntaxError {
+	// Sanity check source supports discarded return values
+	if errs := p.checkDiscardableSource(s.Source); len(errs) > 0 {
+		return errs
+	}
+	// Sanity check at least one return value is bound
+	if !array.ContainsMatching(s.Targets, func(lv LVal) bool { return !isDiscard(lv) }) {
+		return p.srcmaps.SyntaxErrors(s, "at least one return argument must be used")
```
DavePearce: Why enforce this requirement?  Its not necessary.

pkg/zkc/compiler/validate/typing.go
```diff
@@ -300,6 +305,91 @@ func (p *TypeChecker) typeAssignment(s *stmt.Assign[symbol.Resolved], env Variab
 	return append(errors, checkTargets(s, env, p.srcmaps)...)
 }
 
+// typePartialAssignment types an assignment in which one or more targets are
+// the wildcard "_", discarding the corresponding return value.  Discarding is
+// only meaningful for sources producing multiple return values whose types are
+// fixed by a declaration --- i.e. function calls and static memory reads ---
+// so anything else is rejected.  The bound targets are checked positionally
+// against the source's return types.
+func (p *TypeChecker) typePartialAssignment(s *stmt.Assign[symbol.Resolved], env VariableMap, effects bit.Set,
```
DavePearce: I don't really like this function --- its mostly duplicate code.  It would better to just have a few additional checks in typeAssignment().  In particular, because we want to support `_` for other use cases as well.

pkg/zkc/constraints/call_and_memory_lookup.go
```diff
@@ -300,17 +300,26 @@ func emitCallLookup[W vm.Word[W], F field.Element[F]](mod *schema.Table[F, mir.C
...
+		tgtIds []register.Id
 	)
-	//
-	for i := range tgtIds {
-		tgtIds[i] = register.NewId(uint(i))
+	// A discarded return binds no caller register, so its (source, target)
+	// column pair is simply omitted from the lookup.  This remains sound: the
+	// callee-side output column is still fully constrained by the callee's own
+	// table, and nothing on the caller side depends on its value.
+	for i, id := range append(append([]register.Id{}, args...), returns...) {
```
DavePearce, line 332: I mean ... really?

DavePearce: Haha, I see this monstrosity was already in the code base :(

pkg/zkc/constraints/call_and_memory_lookup.go
```diff
@@ -357,18 +366,27 @@ func emitMemoryLookup[W vm.Word[W], F field.Element[F]](mod *schema.Table[F, mir
...
+		tgtIds []register.Id
 	)
-	//
-	for i := range tgtIds {
-		tgtIds[i] = register.NewId(uint(i))
+	// A discarded data line of a (static) read binds no accessor register, so
+	// its (source, target) column pair is simply omitted from the lookup.  This
+	// remains sound: the table row is fixed at compile time, and nothing on the
+	// accessor side depends on the omitted column's value.
+	for i, id := range append(append([]register.Id{}, address...), data...) {
```
DavePearce, line 440: Seems like we could have a function for this, since its repeated for both function calls and memory.

pkg/zkc/constraints/translator.go
```diff
@@ -96,7 +96,7 @@ func translateStaticMemory[W vm.Word[W], F field.Element[F]](_ schema.ModuleId,
 		outputs = toRegisters(m.DataRegisters())
 		// Convert the static contents from words into field elements.
 		contents     = toFieldElements[W, F](m.StaticContents())
-		paddedHeight = util_math.NextPowerOfTwo(uint(len(contents)))
+		paddedHeight = util_math.NextPowerOfTwo(m.StaticHeight())
```
DavePearce: This confused me.  Ok, I see its a bug fix though.

pkg/zkc/vm/internal/bytecode/bytecode.go
```diff
@@ -44,6 +45,15 @@ const (
 // RegisterId just provides a convenient alias to make the code more readable.
 type RegisterId = uint16
 
+// DISCARD is a pseudo register id marking a discarded ("_") return value in a
+// call's return list, or a discarded data line in a static memory read.  It
```
DavePearce: I think the first sentence is enough.  The rest is not very helpful.

pkg/zkc/vm/internal/bytecode/util.go
```diff
@@ -77,6 +82,25 @@ func RegisterToString[W word.Word[W]](reg RegisterId, env Environment[W]) string
 // Helpers
 // ============================================================================
 
+// boundRegisters filters any DISCARD entries out of the given register list,
+// leaving only registers actually bound.  The original list is returned
+// (unchanged) when it contains no DISCARD entries.
+func boundRegisters(regs []RegisterId) []RegisterId {
```
DavePearce, line 88: This is a filter operation.  You can replace with just this:

```
return array.Filter(regs, func(r RegisterId) bool { return r != vm.DISCARD })
```

Personally, I would inline this into the two call sites of `boundRegisters()` and avoid the function entirely.

DavePearce: The key here is that the line above tells you much more than this:

```
return boundRegisters(reg)
```

pkg/zkc/vm/internal/gogen/emit_call.go
```diff
@@ -127,6 +134,16 @@ func (g *generator) emitCall(c *code, fn *descFunction, x *bytecode.Call[word.Ui
 		targets := []string{}
 
 		for _, r := range rets {
+			// A discarded return becomes a blank (one per tuple slot).
+			if !r.bound {
+				targets = append(targets, "_")
+				if r.outWide {
```
DavePearce, line 140: Ok, its going double duty to generate `_, _ = `.  Makes sense.

pkg/zkc/vm/internal/gogen/emit_call.go
```diff
@@ -147,6 +164,16 @@ func (g *generator) emitCall(c *code, fn *descFunction, x *bytecode.Call[word.Ui
 		vals := make([]operand, len(rets))
 
 		for i, r := range rets {
+			// A discarded return becomes a blank (one per tuple slot).
+			if !r.bound {
+				tmps = append(tmps, "_")
```
DavePearce, line 169: Hmmm, a lot duplication going on here.  This gogen code is already pretty ugly though :(

DavePearce: Hmmmm "pretty" is an understatement.  This function is 170 lines long.  The "Power of Ten" rules recommend at most 60 lines!!  Anyway, that is an issue for another day.

pkg/zkc/vm/internal/interpreter/encoding/call.go
```diff
@@ -37,7 +38,39 @@ func Call[W word.Word[W]](pc uint32, p *bytecode.Call[W], env Environment[W]) (c
 	// Encode enter
 	codes = append(codes, encodeEnter(pc, offset, width, p.Arguments)...)
 	// Encode leave
-	return append(codes, encodeLeave_n(p.Returns)...)
+	return append(codes, encodeLeave_n(denseBindings(p.Returns))...)
+}
+
+// denseBindings rewrites a binding list which may contain discarded (DISCARD)
```
DavePearce: This comment is horrific.  Classic "claudesplaining".

DavePearce: That said, the idea of overwriting slots to avoid a special case for `_` is clever.  So, for example, something like `_, a = f(x)` becomes `a,a = f(x)` where the first assignment is just overwritten by the second assignment.  The comment should, at a minimum, use an example like that.

DavePearce: Maybe the name also could be something more like "substituteDiscardedRegisters()" or something which i more helpful.

pkg/zkc/vm/internal/transform/lower_or_xor_and.go
```diff
@@ -56,13 +56,22 @@ func LowerOrXorAnd[W word.Word[W]](program descriptor.Program[W]) descriptor.Pro
 		program.MaxStaticHeight(), append(out, helpers.modules()...)...)
 }
 
-// bitwiseHelperKey identifies an AND/OR/XOR helper module.
+// bitwiseHelperKey identifies a AND/OR/XOR helper module.
```
DavePearce: Yeah, cursorbot is all over this.  It should be `an` when the word is followed by a vowel.

pkg/zkc/vm/internal/transform/lower_or_xor_and.go
```diff
@@ -90,7 +99,7 @@ func (p *bitwiseHelpers[W]) modules() []descriptor.Module[W] {
 }
 
 func helperName(key bitwiseHelperKey) string {
-	return fmt.Sprintf("$bit_%s_u%d", bitwiseOpName(key.op), key.width)
+	return fmt.Sprintf("$bit_AndOrXor_u%d", key.width)
```
DavePearce: I don't like this name because we have, thus far, followed Rust convention and consistently used `lower_snake_case`.  I think `bit_xoa` is absolutely fine, or some variation (previously we had `bit_xoan` in zkASM, hence my suggestion).

pkg/zkc/vm/bytecode.go
```diff
@@ -59,6 +59,13 @@ type Register[W Word[W]] = descriptor.Register[W]
 // bytecode RegisterId).
 type RegisterId = bytecode.RegisterId
 
+// DISCARD is a pseudo register id marking a discarded ("_") return value in a
```
DavePearce: Again, first sentence is enough.

testdata/zkc/unit/partial_call_05.zkc
```diff
@@ -0,0 +1,26 @@
+pub input data(address:u16) -> (byte:u16)
+
+// A two-column static table whose hi column is 16 times its lo column, so
+// that a value read from one column predicts the other.
+static tbl(address:u16) -> (lo:u16, hi:u16) {
+    (0x0a, 0xa0),
```
DavePearce, line 6: I like the tuple initialisers --- its a nice touch.

DavePearce (changes requested):
Overall, I think this looks pretty good.  I have requested some changes, but they are mostly pretty minor.

pkg/zkc/constraints/translator.go
letypequividelespoubelles: yep, we don't count the same way the static table length in ast and in air model ... :(

pkg/zkc/compiler/parser/parser.go
letypequividelespoubelles: indeed

pkg/zkc/compiler/validate/typing.go
letypequividelespoubelles: It's not necessary, but I see it as a "garde-fou". I don't see any use case where we would like `_ = f(x)`

pkg/zkc/compiler/validate/typing.go
```diff
@@ -262,44 +262,109 @@ func (p *TypeChecker) typeMemory(c *decl.ResolvedMemory) []source.SyntaxError {
...
 		_, errors = p.typeExpression(nil, s.Source, env, effects)
-		//
 		return errors
 	}
-	// Type check left-hand side
-	for i, lval := range s.Targets {
+	// Partial assignment: one or more targets are the wildcard "_".
+	// Discarding is only meaningful for sources whose return types are fixed by
+	// a declaration (function calls and static memory reads).
+	if array.ContainsMatching(s.Targets, isDiscard) {
```
DavePearce, line 274: Oh, I love this so much.  I can only hope that Claude is solely responsible ;)

DavePearce (approved):
Overall, looks good.  The change merging `typePartialAssignment()` into `typeAssignment` was not done well, but otherwise looks good.

PR #2193, opened 2026-08-27
fix: honour already provided source files when compiling

ivokub:
When we do:
```
compiler.Compile(field.KOALABEAR_16, cfg.GetMaxStaticHeight(), srcFiles...)
```
we can already provide all the source files. However, in the pre-PR implementation we still searched the file system for the file when looking for files to link even if it was already provided. This prevents for example embedding the arithmetization into a binary (through an archive or `fs.FS` implementation for example).

This PR fixes the case and makes the source file searcher to skip loading files from file system if it is already provided in a call to `compiler.Compile`.


DavePearce (approved):
LGTM

Issue #2000, opened 2026-07-17
feat: support constant bytecode operands


DavePearce:
_(see also #1838)_

# Summary

Certain bytecode instructions do not accept constants as operands, including `Call`, `MemRead`, `MemWrite`, `Bitwise`, `Cat`, `DivRem`, `Intrinsic` and `SkipIf`.  This is causing a large amount of overhead and the purpose of this issue is to eliminate that.  A key constraint here is that (aside from `SkipIf`) these instructions require registers as operands at the AIR level.  There are two plausible approaches: _constant registers_ (see #1838) and _constant operands_ (see below).  One point to note is that, whilst the constant operands approach still requires constant registers at the AIR level, it does offer a more friendly interface.

## Approach (Constant Operands)

In this approach, the goal is to allow the operands of a e.g. `Call` bytecode to be either a `RegisterId` (as they currently are) or a constant `W`.  To make this meaningful, we change the signature of these bytecodes from this:

```
type Call[W word.Word[W]] struct {
	// address of target function
	Target ModuleId
	// Arguments are caller-frame registers copied into callee inputs.
	Arguments []RegisterId
	// Returns are caller-frame registers receiving callee outputs.
	Returns []RegisterId
}
```

To this:

```
type Call[W word.Word[W]] struct {
	// address of target function
	Target ModuleId
	// Arguments are caller-frame registers copied into callee inputs.
	Arguments []Operand[W]
	// Returns are caller-frame registers receiving callee outputs.
	Returns []Operand[W]
}
```

Here, an `Operand[W]` is either a `RegisterVector` or a `ConstantVector` (i.e. which wraps a `[]W`).

**Challenges**

* (**Interpreter**) this needs to be updated to support encodings of these instruction forms which admit constants.  Also, certain instructions will need to include a bitwidth for any constant operands (e.g. shift, divrem, 
* (**Constraint Generator**) this needs to be updated to allocate constant registers for any constant operands (whilst ensuring the minimal number of registers are created), whilst also ensuring that these come at the end of the enclosing module's register file (i.e. to ensure these registers are not filled when tracing, including for those `Assignment` instances currently in play).
* (**Register Splitting**) this needs to be updated to split constant operands, and ensure alignment of such operands (as it currently does for register operands).
* (**Constraint Generator**) For `SkipIf` there are various changes throughout which need to be made to support a constant operand, though this instruction could be handled separately.
[...]

Issue #1969, opened 2026-07-07
feat: prevent tracing for constant registers


DavePearce:
# Summary

Constant registers should not be traced, but currently they are (since this is the only way they can work).  The proposed solution is to ensure constant registers always have the highest register ids in the module, such that a trace row is a consecutive sequence of cells (i.e. no constant registers are within the sequence leading to holes, etc).

DavePearce:
Subsumed by #2236

Issue #2187, opened 2026-08-25
feat: reduce number of lowered shift modules


DavePearce:
# Summary

Currently, a reasonable number of shift modules are being created and it seems reasonable that this can be reduced, especially as this is consuming a reasonable number of columns.  Many of these modules have relatively few lines, and hence combining them makes sense.

## Analysis

The list of shift modules generated for the arithmetisation is:

```
                       $bit_shl_u64_u2 |      25
                       $bit_shl_u64_u3 |      25
                       $bit_shl_u64_u4 |      25
                       $bit_shl_u64_u6 |      22
                       $bit_shl_u64_u5 |      21
                       $bit_shl_u64_u1 |      19
                       $bit_shr_u32_u2 |      15
                       $bit_shr_u32_u3 |      15
                       $bit_shr_u32_u4 |      15
                       $bit_shr_u32_u5 |      12
                       $bit_shr_u32_u1 |      11
                       $bit_shl_u32_u2 |      15
                       $bit_shl_u32_u3 |      15
                       $bit_shl_u32_u4 |      15
                       $bit_shl_u32_u1 |      11
                       $bit_shr_u64_u2 |      27
                       $bit_shr_u64_u3 |      27
                       $bit_shr_u64_u4 |      27
                       $bit_shr_u64_u1 |      21
                       $bit_shr_u64_u5 |      20
                       $bit_shr_u64_u6 |      20
                       $bit_shl_u21_u1 |      10
                       $bit_shl_u32_u5 |      13
                       $bit_shr_u64_u7 |      13
                        $bit_shl_u5_u2 |      10
                        $bit_shl_u5_u1 |       7
```

Next to each module, the number of columns is listed.

## Approach

There are two obvious approaches:

1. (**Combine SHL/SHR**) Given the similarity of the two modules, it seems pretty reasonable to combine them together with a "selector" argument used to determine the kind of shift.
2. (**Single Decomposition**) Currently, each different type is being decomposed (e.g. we have `bit_shl_u64_u3` and `bit_shl_u32_u3`) even though the underlying register size is `u16`.  In contrast, for the bitwise operators, we see only `bit_and_u16`, `bit_or_u16`, etc.
[...]

Issue #2108, opened 2026-08-05
feat: support deduplication for tracing


letypequividelespoubelles:
# Summary

Currently, the tracer does not deduplicate traces at all --- even when it is permitted to do so.  This leads to additional lines in the generate trace which could be eliminated.  However, care needs to be taken to ensure deduplication is done safely.

## Approach

Any function which does not have memory effects can have its trace deduplicated.  For example, a function like this:

```
fn add(x:u32,y:u32) -> (r:u32) {
  r = x + y
}
```

If this function is called with the same arguments twice (e.g. `add(1,2)`) then, currently, there will be two identical rows in the generated trace for the `add` module.

**Multi-line functions**.  To deduplicate multi-line functions, we must always work at the level of whole frames.  That is, we must identify entire frames which are identical and remove them.  This can complicate the process.

## Notes

* (**zkASM**).  Deduplication was done here during the trace propagation phase using a cache of function parameters / returns.  Thus, when a function was called with the same parameters / returns these were simply returned from the cache rather than being re-executed.

Its not clear where the right place to implement deduplication is.  There are several options: (1) during tracing where rows are simply dropped rather than being appended; (2) during trace post-processing by doing some kind of sorting and deduplication; (3) during trace execution actually apply "short-circuiting" using a cache to avoid re-executing functions entirely.  Notes:

* **Option (2)** would be fine for one-line functions, but more challenging for multi-line functions.
* **Option (3)** could save time overall if there are a large number of duplication executions (e.g. for bitwise operators).  However, if there aren't that many --- then it hurts overall performance.  The experience with trace propagation in zkASM was that a cache using a suitable read/write lock can be pretty efficient.

To offset the overhead of option (3) it might be possible for the user to specify with an annotation whether or not [function calls should be "memoized".](https://en.wikipedia.org/wiki/Memoization).  This would allow us to specify those which have high amounts of duplication in practice.

DavePearce:
Stats based on a "tiny block":

```
                                       | columns |    lines |   unique | bitwidth |      cells |   nonzero |      bytes |
                                  main |     299 | 16777216 | 16777216 |     2909 | 5016387584 | 847126598 | 7549747200 |
            process_I_type_instruction |     361 |  8388608 |  8388608 |     2979 | 3028287488 | 447055125 | 4102029312 |
                           interpreter |     141 | 16777216 | 16777216 |     1577 | 2365587456 | 452694930 | 3758096384 |
[...]

PR #2195, opened 2026-08-28
perf: split keccak benchmark into narrow and wide variants

gbotrel:
Two implementations of the same keccak, side by side. We keep the winner and delete the loser.

Both: fused round (θ ρ π χ ι in one pass), constant rotations as `::` splits instead of runtime shifts, no integer division, and the padding written into ram so the last block is absorbed by the same code as every other block.

- **narrow** — state stays in `keccak_state` memory between rounds. 50 memory accesses per round.
- **wide** — state stays in registers, passed through calls. No memory in the permutation, but 50-value call signatures.

KOALABEAR_16, 4,236-byte message. Columns / lookups / complexity are function-module totals from `zkc compile --stats`; cells from `zkc trace --stats`.

| | columns | lookups | complexity | cells |
|---|---:|---:|---:|---:|
| old | 3645 | 3617 | 195,288 | 35.9M |
| narrow | 3276 | 3389 | 71,479 | 7.4M |
| wide | 3343 | 3454 | 92,861 | 6.2M |

Both beat old on every metric. narrow and wide are within 2% on columns; wide wins cells by 16%, narrow wins complexity by 23%.

wide looks like the keeper. narrow's `keccak_round` costs 1386 columns against wide's 1184, and that 202 difference is pure bookkeeping for the 50 memory accesses — the arithmetic is identical. narrow also drags 1.5M cells of `keccak_state` traffic that wide does not have.

### Why not a `rotl` function

Tried it. `keccak_round` drops 1184 → 929 columns, but the `$bit_shl` / `$bit_shr` / `$bit_or` tables it pulls in give most of that straight back: 3343 → 3288 columns total (−1.6%), while cells go 6.2M → 17.5M (2.8×). Even if those tables were free the ceiling is −7% columns. Not worth it yet — revisit after #2000 / #1969.

### Notes

Both variants write the padding into ram in `[msg_length, block boundary)`, so the caller must leave that range free.

`ParallelTracing` is dropped from the keccak tests: sharded tracing of read-write memories violates `active_monotony` (see the `util.Config.ParallelTracing` docstring). Pre-existing on main.

Verified: all 164 vectors on both variants, `trace --check` clean, `Test_ZkcBench_Keccak{Narrow,Wide}` green, `make zkc-lint` clean.


DavePearce:
The cell reduction for both wide and narrow implementations looks great, but they are not attacking the problem of columns.  Overall, I am a little surprised the `narrow` implementation does not do better.  Looking at some specifics, the implementation for `narrow` looks odd:

```
var r3h:u3, r3l:u61
r3h::r3l = a02 ^ d0
var b21:u64 = r3l::r3h
```

This is introducing a lot of temporary columns, instead of using a separate `rotl` function.  Also, `absorbing_phase` can be further improved.

DavePearce:
With improvements to keccak the wide implementation does look like the winner to me.  The biggest challenge is that we are optimising for something which doesn't exist yet.

We can further reduce the number of columns by e.g reintroducing the rotl function.  But, then the cell count goes from this:

```
Cells |    6.1M |
             Cells (raw) | 6143716 |
     Columns (1..8 bits) |     822 |
    Columns (9..16 bits) |    3028 |
   Columns (17..32 bits) |       0 |
  Columns (33..128 bits) |       0 |
 Columns (129..256 bits) |       0 |
        Columns (native) |      50 |
------------------------------------------------------------------------------
                 | columns |  lines | bitwidth |   cells | nonzero |   bytes |
    $bit_xor_u16 |      10 | 262144 |       97 | 2621440 | 2328733 | 3407872 |
    $bit_and_u16 |      10 | 131072 |       97 | 1310720 |  750469 | 1703936 |
    keccak_round |    1184 |   1024 |    16250 | 1212416 |  834025 | 2240512 |
        keccak_f |     318 |   1024 |     4946 |  325632 |  244513 |  640000 |
       copyInput |     118 |   2048 |      889 |  241664 |   47329 |  319488 |
          keccak |    1294 |    128 |    17448 |  165632 |   30837 |  301696 |
```
To this:
```
Cells |    17.4M |
             Cells (raw) | 17416810 |
     Columns (1..8 bits) |      810 |
    Columns (9..16 bits) |     2987 |
   Columns (17..32 bits) |        0 |
  Columns (33..128 bits) |        0 |
 Columns (129..256 bits) |        0 |
        Columns (native) |       50 |
------------------------------------------------------------------------------
                 | columns |  lines | bitwidth |   cells | nonzero |   bytes |
    $bit_xor_u16 |      10 | 262144 |       97 | 2621440 | 2328733 | 3407872 |
    $bit_and_u16 |      10 | 131072 |       97 | 1310720 |  750469 | 1703936 |
     $bit_or_u16 |      10 | 131072 |       97 | 1310720 |  654527 | 1703936 |
    keccak_round |     929 |   1024 |    13434 |  951296 |  653456 | 1795072 |
 $bit_shr_u64_u2 |      27 |  32768 |      309 |  884736 |  255557 | 1507328 |
 $bit_shr_u64_u3 |      27 |  32768 |      311 |  884736 |  275831 | 1507328 |
```
Which looks bad ... but its actually not really.

Specifically, because the various `$bit_` related functions are not (yet) eliminating duplicates.  When this is implemented, they will look quite different.  We'd end up with a comparison more like this:

```
Cells |    6.1M |
...
------------------------------------------------------------------------------
                 | columns |  lines | bitwidth |   cells | nonzero |   bytes |
    keccak_round |    1184 |   1024 |    16250 | 1212416 |  834025 | 2240512 |
        keccak_f |     318 |   1024 |     4946 |  325632 |  244513 |  640000 |
       copyInput |     118 |   2048 |      889 |  241664 |   47329 |  319488 |
          keccak |    1294 |    128 |    17448 |  165632 |   30837 |  301696 |
             ram |      28 |   4096 |      297 |  114688 |   23321 |  184320 |
```
then becomes:
```
Cells |    ???M |
------------------------------------------------------------------------------
                 | columns |  lines | bitwidth |   cells | nonzero |   bytes |
    keccak_round |     929 |   1024 |    13434 |  951296 |  653456 | 1795072 |
        keccak_f |     318 |   1024 |     4946 |  325632 |  244513 |  640000 |
       copyInput |     118 |   2048 |      889 |  241664 |   47329 |  319488 |
          keccak |    1294 |    128 |    17448 |  165632 |   30837 |  301696 |
             ram |      28 |   4096 |      297 |  114688 |   23321 |  184320 |
```
And the only difference is actually in `keccak_round`.

But the number of columns for `keccak_round` falls from `1184` to `929`.  I think the improvement is not as much as it could be, because there are a lot of constant columns being added.  Again, with support implemented for this, constant columns should be actually free.

Likewise, we'll be expecting improvements to the `$bit_shr_XXX` modules as well, since these should be benefitting from reference tables for the small shifts.

Then, relevant outstanding issues are:

- https://github.com/LFDT-Lineth/zkc/issues/2108
- https://github.com/LFDT-Lineth/zkc/issues/2187
- https://github.com/LFDT-Lineth/zkc/issues/2000 / https://github.com/LFDT-Lineth/zkc/issues/1969

PR #2197, opened 2026-08-29
chore: reorg tests (again)

### LINKED ISSUE #2196: chore: reorganise legacy tests (again)
# Summary

A previous PR (of mine) reorganised the tests to prohibit bare `accepts` files (e.g. `basic_01.accepts`) and, instead, to force them to specify their field (e.g. `basic_01.koalabear_16.accepts`).  This was helpful in that it broke the original dependency of base `accepts` files actually defaulting to the `BLS12_377` field. However, overall, this is quite wasteful in terms of file size and its better to have bare `accepts` files which (now) pass for all fields, and then field-specific tests only for those vectors which require it.

DavePearce:
Not sure why I am going back and forth on this, but have decided to go back to the original test organisation.  This is where a bare "accepts" file is run for all fields, whilst a field-specific accepts file is only run for the specific field.


letypequividelespoubelles (approved):
LGTM

PR #2202, opened 2026-09-01
feat: enable fast mode for sharded tracing

### LINKED ISSUE #2198: feat: enable fast mode for sharded tracing
# Summary

Currently, sharded tracing uses the "tracing program" to do the initial checkpointing run.  However, this is relatively inefficient compared with using fast mode.  However, enabling fast mode for this purpose is non trivial.

## Notes

Some things to take into consideration:

* (**Agnostic Checkpoints**).  Making the checkpoints "word agnostic" is a sensible approach.  To make this work requires an efficient, allocation-free approach to `Pack()/Unpack()` checkpoint pages to/from the serialised form.  Previous attempts have failed because this allocates too much memory, and was not efficient.
* (**ROMs / SROMs**).  These should not be included in the checkpoint itself. 
* (**Call Stack**).  Since checkpoints are always made at `#[global]` points, we actually do not need to record the stack frame at all.  Instead, we just need to know the function we are continuing, and the state of memory.
* (**Copy-On-Write**). Using COW pages within the interpreter could offer relatively cheap checkpoints, further reducing their cost.

DavePearce:
This enables fast mode execution when sharded tracing.  Specifically, fast mode is used during the initial checkpointing run before tracing begins proper.


letypequividelespoubelles (approved):
lgtm

Issue #2160, opened 2026-08-18
Sharded RAM consistency v2


OlivierBBB:
# RAM, RAM_FINALIZER and RAM_FINALIZATION_RECORD modules

This is a follow-up issue to #2123.

`RAM` and `RAM_PERM` check shard level consistency.
`RAM_FINALIZER` and `RAM_FINALIZATION_RECORD` are disconnected from every other module (though `RAM_FINALIZER` is connected to the perm. bus) and the only connection required between them is the `RAM_FINALIZER ↪ RAM_FINALIZATION_RECORD` lookup.

## RAM

The RAM module looks like so

```
| ADDRESS | VALUE_READ | TIMESTAMP_READ | VALUE_WRITTEN | TIMESTAMP_WRITTEN | IS_READ |
|:-------:+:----------:+:--------------:+:-------------:+:-----------------:+:-------:|
|    a    |            |                |       W       |         tk        |    0    |
|    a'   |            |                |       W'      |        tl'        |    0    |
|    a"   |            |                |       W"      |        tm"        |    1    |
```

```rust
// RAM module constraints
TIMESTAMP_WRITTEN[i] = 1 + TIMESTAMP_WRITTEN[i-1]
TIMESTAMP_READ[i] < TIMESTAMP_WRITTEN[i]
if IS_READ[i] = 1 Then
   VALUE_WRITTEN[i] = VALUE_READ[i]
```

**Edit.** Discussion this with @letypequividelespoubelles highlighted that we can get away with fewer temporal columns. All we really need in the temporal realm is
```rust
// temporal columns
TIMESTAMP_WRITTEN

// temporal constraints
If EXEC[i-1] = 1 Then EXEC[i] = 1
If EXEC[i-1] = 1 Then TIMESTAMP_WRITTEN[i] = 1 + TIMESTAMP_WRITTEN[i-1]

// permutation constraint
(EXEC, TIMESTAMP_WRITTEN) is a permutation of (EXEC, perm_TIMESTAMP_WRITTEN)

// optionally
If EXEC[i] = 0 Then TIMESTAMP_WRITTEN[i] = 0
```

and we move the previous constraints to the permuted realm:
```rust
// RAM module constraints
perm_TIMESTAMP_WRITTEN[i] = 1 + perm_TIMESTAMP_WRITTEN[i-1]
perm_TIMESTAMP_READ[i] < perm_TIMESTAMP_WRITTEN[i]
[...]

PR #2205, opened 2026-09-07
feat: temporal timestamp column for RAM, drop the unused FINL phase

OlivierBBB:
Part 1/4 of #2160 / #2206 (plan and prover-side assumptions in the #2160 comment).

Two changes to the RAM module:

1. **Temporal timestamp column.** `$temporal_ts_k` counts the shard's clock: on consecutive real rows `TEMPORAL_TS[i] = TEMPORAL_TS[i-1] + 1` (the interpreter clock ticks once per access). The tracer fills it with `first stamp + row index`. Today rows are still traced in access order, so it equals `TIMESTAMP_WRITTEN`; #2206 will trace rows in `(ADDRESS, TIMESTAMP_WRITTEN)` order and tie the two columns with a permutation, so `TIMESTAMP_WRITTEN` is deliberately *not* constrained to be chronological.

2. **FINL phase removed.** The never-populated init/finalization phase is gone (`FINL`, `ADDR_DELTA`, `ADDR_CARRY`, all `finl_*` constraints, `exec_and_finl_are_binary_exclusive`; `active_monotony` becomes `exec_monotony`). Finalization becomes a separate module in part 4. `EXEC` stays and now just means "real row, not padding".

Review points:

- Guard is `EXEC[i-1] ∧ EXEC[i]`: the first row after padding — hence a shard's first row — is unconstrained, and the clock continues across shards (first stamp of a shard = checkpoint clock + 1; stamp 0 stays reserved for never-touched cells).
- Multi-limb `+1`: `multiLimbIncrement` with a nil delta. The carries need their own family `$temporal_ts_carry_k` (n−1 one-bit columns, empty on single-limb fields) because `$ts_carry_k` witnesses the same-row sum `TS_READ + 1 + TS_DELTA`. The one-hot `$at_flag` scheme used for ROM/WOM addresses is equivalent but costs L columns and ~L²/2 equalities instead of L−1 columns and L equations.
- New columns are appended after `EXEC_READ`, so the ids used by the caller→RAM lookup are unchanged; removing FINL shifts the ids after `IS_WRITE`, but all three sites (`computeRamLayout`, `newRamTraceLayout`, `emitRamLookup`) derive them from the same layout.
- Checks: `make lint` clean; `Test_ZkcUnit|Test_ZkcMixed|Test_ZkcBench` pass on all fields (incl. the sharded RAM cases).


DavePearce (approved):
LGTM

PR #2207, opened 2026-09-09
perf: reduce numbers of shift modules

### LINKED ISSUE #2187: feat: reduce number of lowered shift modules

letypequividelespoubelles:


pkg/zkc/vm/internal/transform/lower_bitwise.go
```diff
@@ -107,12 +122,40 @@ func lowerBitwiseShlShr[W word.Word[W]](
...
+	default:
+		panic("unsupported opcode")
+	}
+	//
+	code := []Bytecode[W]{bytecode.CallFun[W](uint16(id),
+		[]bytecode.RegisterId{b.Left, amount}, returns)}
+	// Truncate by destructing off the high bits (little-endian, so the target
+	// takes the low width bits).
+	if truncate {
+		high := registers.Allocate("shl_res_high", util.Some(maxWidth-width))
```
amkCha, line 153: I don't know what impact it has, but we can't discard this computed register afterwards ? As it's the part we won't keep

amkCha (approved):

pkg/zkc/vm/internal/transform/lower_bitwise.go
letypequividelespoubelles: Not for now. The only ones we can't discard are the one used to cut intra u16 word. But all the highs u16 full words could be discarded. But it's a more global perf issue we have.

PR #2209, opened 2026-09-14
fix: explicit ordering constraint for LowerSwitch

### LINKED ISSUE #2208: fix: mark `LowerSwitch` to run after `FactorSkipConditions`
# Summary

The `LowerSwitch` transform should run after `FactorSkipConditions`, and should be marked as an explicit constraint.

## Minor

Can update the comment for `factorableSkips()` in `factor_skip_conditions` like so:

```
// Nothing to factorize if the body of the skip is a bit equality like b = x == 0 ? 1 :0.
// Note that as we lowerSwitch later, this pattern can't arise from lowerSwitch, but only
// directly from .zkc program or other lowering steps (FactorLimbEqualities ...)
```

DavePearce:
This adds an explicit constraint to ensure that LowerSwitch always comes after FactorSkipConditions.


letypequividelespoubelles (approved):
LGTM

PR #2214, opened 2026-09-15
feat: remove notion of "Computation"

### LINKED ISSUE #2107: feat: tidy `ComputedRegister`
# Summary

The `ComputedRegister` type is used in a small number of places, but its requirements are lower than before.  Specifically, it was previously the case that a computed register could be subject to splitting, but this is no longer the case.  Instead, the efficiency of a computed register could be dramatically improved by requiring it to target only a single register, and avoiding a `Computation[word.BigEndian]`

DavePearce:
The notion of a computation was previously required to assist computed register assignments in the presence of field agnosticity, and register splitting.  However, the role of such assignments has been diminished to the point where there is no benefit from this concept.  Hence, it removed.


letypequividelespoubelles (approved):
It was fun using this :)

